PR #13, opened 2020-04-08
fix: tests for ApiGatewayHandler

axthosarouris:


src/main/java/nva/commons/hanlders/ApiGatewayHandler.java
```diff
@@ -72,6 +71,7 @@
      *
      * @param iclass The class object of the input class.
      */
+    @JacocoGenerated
```
axthosarouris, line 74: This cannot be tested. It necessary for the aws lambda

src/main/java/nva/commons/hanlders/ApiGatewayHandler.java
```diff
@@ -85,13 +85,13 @@ public ApiGatewayHandler(Class<I> iclass) {
...
         this.environment = environment;
-        this.allowedOrigin = environment.readEnv(ALLOWED_ORIGIN_ENV);
-        this.additionalSuccessHeadersSupplier = () -> Collections.emptyMap();
+        this.additionalSuccessHeadersSupplier = Collections::emptyMap;
     }
 
     private void init(OutputStream outputStream, Context context) {
         this.outputStream = outputStream;
         this.logger = context.getLogger();
+        this.allowedOrigin = environment.readEnv(ALLOWED_ORIGIN_ENV);
```
axthosarouris, line 94: Moved because the tests "ApiGatewayHandler has a constructor with input class as only parameter" cannot pass otherwise

PR #16, opened 2020-04-10
fix: Try improvements and testing

axthosarouris:
New feature:
* orElse() function

Tests:
* Try/Success/Failure tests

### orElse functionality
 * input: a function that takes as input a Failure object. 
 * output: If the called object is a Success, it ignores the input and returns the value that it holds. If the called object is a Failure then it applies the function on itself and returns a value depending on the defined function or throws an Exception.

* orElse() examples:
 See tests.

### DIFF
--- a/src/test/java/nva/commons/utils/attempt/FailureTest.java
+++ b/src/test/java/nva/commons/utils/attempt/FailureTest.java
@@ -43,7 +43,7 @@ public class FailureTest {
     public void orElseThrowsTheSpecifiedException() {
         Executable action =
             () -> Try.of(sample)
-                     .map((Integer i) -> illegalAction(i, NESTED_EXCEPTION_MESSAGE))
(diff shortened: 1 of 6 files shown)


src/test/java/nva/commons/utils/attempt/FailureTest.java
```diff
@@ -0,0 +1,140 @@
...
+    private static final Integer DEFAULT_VALUE = 100;
+
+    private final Integer sample = 1;
+
+    @Test
+    @DisplayName("orElseThrow throws illegalStateException for null argument")
+    public void orElseThrowsIllegalStateExceptionForNullArgument() {
+
+        Executable action =
+            () -> Try.of(sample)
```
axthosarouris, line 34: Try.of(value) returns a new Success with the value.

PR #17, opened 2020-04-10
fix: ignore readEnvOpt from test coverage

axthosarouris:


src/main/java/nva/commons/utils/Environment.java
```diff
@@ -14,6 +14,7 @@
      * @param variableName the Env variable name.
      * @return the value of the variable or an empty Optional if the variable does not exists.
      */
+    @JacocoGenerated
```
axthosarouris, line 17: Cannot test for empty values consistently when reading from an actual environment. Jacoco complains for the fact that we do not test both outcomes of the filter

PR #19, opened 2020-04-14
fix: exceptions tests

axthosarouris:


src/test/java/nva/commons/exceptions/ApiGatewayExceptionTest.java
```diff
@@ -54,11 +56,20 @@ public void getStatusCodeThrowsIllegalStateExceptionWhenStatusCodeHasNotBeenSet(
     @Test
     @DisplayName("statusCode returns Integer and not int")
```
axthosarouris, line 57: Not set is "null" and not a specific value (0)

PR #25, opened 2020-04-15
feat: TestContext

axthosarouris:


src/main/java/nva/commons/utils/TestContext.java
```diff
@@ -0,0 +1,75 @@
+package nva.commons.utils;
```
axthosarouris, line 1: Using Mockito here would mean a dependency on Mockito version for all projects.

PR #27, opened 2020-04-16
Added full pathParameters map to RequestInfo

janbj:


src/main/java/nva/commons/handlers/RequestInfo.java
```diff
@@ -12,10 +11,11 @@
     public static final String PROXY_TAG = "proxy";
     private Map<String, String> headers;
     private String path;
-    private String proxy;
+    private Map<String,String> pathParameters;
```
axthosarouris, line 14: Which fields do we need exactly from the map?

janbj: The fields we need, it can be anything. Example 
`              parameters:
                - in: path
                  name: identifier
                  type: string
                  format: uuid
                  required: true
                  description: UUID identifier of the Publication to modify.`

PR #30, opened 2020-04-17
fix: Void class and better Environment testing

axthosarouris:


src/main/java/nva/commons/handlers/Void.java
```diff
@@ -0,0 +1,4 @@
+package nva.commons.handlers;
+
+public final class Void {
```
brinxmat, line 3: Is there a reason why we don't use Java's native Void type?

axthosarouris: I didn't know about it. Fixed

PR #33, opened 2020-04-22
Added RequestInfo.requestContext

janbj:


src/main/java/nva/commons/handlers/RequestInfo.java
```diff
@@ -15,6 +15,7 @@
     private Map<String, String> pathParameters;
     @JsonProperty("queryStringParameters")
     private Map<String, String> queryParameters;
+    private Map<String, Object> requestContext;
```
axthosarouris, line 18: What can the object be?

PR #34, opened 2020-04-23
RequestInfo map-like structures get initialized as empty non null structures.

axthosarouris:
RequestInfo map-like structures get initialized as empty non null structures when values are null or missing

### DIFF
--- a/src/main/java/nva/commons/handlers/RequestInfo.java
+++ b/src/main/java/nva/commons/handlers/RequestInfo.java
@@ -4,7 +4,6 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.HashMap;
 import java.util.Map;
-import nva.commons.utils.JacocoGenerated;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class RequestInfo {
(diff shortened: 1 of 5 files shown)


src/main/java/nva/commons/handlers/RequestInfo.java
```diff
@@ -19,50 +19,98 @@
...
     public RequestInfo() {
         this.headers = new HashMap<>();
         this.pathParameters = new HashMap<>();
         this.queryParameters = new HashMap<>();
+        this.requestContext = JsonUtils.jsonParser.createObjectNode();
     }
 
-    @JacocoGenerated
+    /**
+     * headers.
```
brinxmat, line 33: Adequate?

src/main/java/nva/commons/handlers/RequestInfo.java
```diff
@@ -19,50 +19,98 @@
...
 
-    @JacocoGenerated
+    /**
+     * Set the pathParameters map. Null input results to an empty map.
+     *
+     * @param pathParameters the pathParameters map.
+     */
     public void setPathParameters(Map<String, String> pathParameters) {
-        this.pathParameters = pathParameters;
+        if (pathParameters == null) {
```
brinxmat, line 87: I still find Objects::isNull easier to read

PR #37, opened 2020-04-27
Np 711 Log4j2 logger and tests

axthosarouris:


build.gradle
```diff
@@ -81,6 +90,14 @@ checkstyle {
     showViolations = true
 }
 
+//For usage in a shadow jar.
```
axthosarouris, line 93: This is for documentation reasons, in case someone wants to use shadow jars. I spend a lot of time trying to find the reason why `aws cloudformation deploy` was not working. `sam deploy` is working  fine however.

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
```diff
@@ -91,10 +96,13 @@ public ApiGatewayHandler(Class<I> iclass, Environment environment) {
         this.additionalSuccessHeadersSupplier = Collections::emptyMap;
     }
 
-    private void init(OutputStream outputStream, Context context) {
+    protected void init(OutputStream outputStream, Context context) throws LoggerNotSetException {
         this.outputStream = outputStream;
-        this.logger = context.getLogger();
         this.allowedOrigin = environment.readEnv(ALLOWED_ORIGIN_ENV);
+        if (isNull(logger)) {
```
axthosarouris, line 102: Maybe ApiGatewayHandler to require logger in the constructor

src/main/java/nva/commons/exceptions/LoggerNotSetException.java
```diff
@@ -0,0 +1,21 @@
+package nva.commons.exceptions;
+
+import org.apache.http.HttpStatus;
+
+public class LoggerNotSetException extends ApiGatewayException {
+
+    public static final String WARNING_MESSAGE = "Logger not set for handler:";
```
janbj, line 7: Add a space to the end

src/main/java/nva/commons/exceptions/LoggerNotSetException.java
```diff
@@ -0,0 +1,21 @@
...
+    public static final String DIRECTIONS = "Set logger in handler constructor: logger = "
+        + "LoggerFactory.getLogger(MyHandler.class) ";
+    public static final String META_FORMATTING = "%s%%s%s";
```
janbj, line 10: One '%' too much?

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
```diff
@@ -55,28 +60,28 @@
...
     public static final String ALLOWED_ORIGIN_ENV = "ALLOWED_ORIGIN";
 
-    protected transient String allowedOrigin;
+    protected Logger logger;
 
+    private final transient Class<I> iclass;
     protected final Environment environment;
 
+    protected transient OutputStream outputStream;
+    protected transient String allowedOrigin;
```
janbj, line 71: What happens with this if we create an instance without an Environment reader, empty value in header?

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
```diff
@@ -297,23 +306,24 @@ protected void writeUnexpectedFailure(I input, Exception exception, String reque
...
             inputObject = parseInput(inputString);
 
             O response;
             response = processInput(inputObject, inputString, context);
 
             writeOutput(inputObject, response);
         } catch (ApiGatewayException e) {
-            logger.log(e.getMessage());
-            logger.log(getStackTraceString(e, context.getAwsRequestId()));
+            logger.warn(e.getMessage());
```
janbj, line 321: Just warn, shouldn't this be error?

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
```diff
@@ -324,8 +334,8 @@ private String getStackTraceString(Exception e, String requestId) {
         String causeQueue = CAUSE_PREFIX + createCauseString(e);
         String stackTrace = STACK_TRACE_PREFIX + arrayToStream(e.getStackTrace());
         String suppressed = Optional.ofNullable(arrayToStream(e.getSuppressed()))
```
janbj, line 336: Extract to method

src/main/java/nva/commons/utils/log/LogUtils.java
```diff
@@ -0,0 +1,102 @@
...
+
+    public static final String PATTERN_ATTRIBUTE_NAME = "pattern";
+    public static final String LOG_CLASS_NAME_AND_MESSAGE = "%c{1.}:  %msg";
+    public static final String ROOT_NULL_APPENDER = "NullAppender";
+
+    private LogUtils() {
+    }
+
+    /**
+     * This method sholuld ALWAYS be called before the creation of the object you want to test.
```
janbj, line 33: typo

src/main/java/nva/commons/utils/log/LogUtils.java
```diff
@@ -0,0 +1,102 @@
...
+     *
+     * @param clazz The class of the object under test.
+     * @param <T>   The class of the object under test
+     * @return a {@link TestAppender}
+     */
+    @SuppressWarnings("PMD.CloseResource")
+    public static <T> TestAppender getTestingAppender(Class<T> clazz) {
```
janbj, line 40: This ia a lot of configuration, maybe you can add a reference to the Log4j2 documentation for configuration.

src/main/java/nva/commons/exceptions/LoggerNotSetException.java
axthosarouris: No, I want the ERROR_MESSAGE_PATTERN, to be 
`WARNING_MESSAGE%sDIRECTIONS`, so I have to escape the middle "%" for it to be preserved in the second format command.

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
axthosarouris: You will get an Internal Server Error and a message saying that you are missing the ALLOWED_ORIGIN env variable

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
axthosarouris: I think that if the exception is "expected" we should a return a warning. E.g. if the user sends invalid information then it is not a system error and we should log it as a warning. The user will get an error code and will send another query. If however, it is something that we did not expect (i.e. bug) then we should log it as an error, because it is an error in the system.

build.gradle
brinxmat: Do you need to comment the code out, just add a comment to the effect that the task can be run in the event that you need to build a shadow jar — it doesn't affect the rest of the code and you've included the plugin anyway

PR #45, opened 2020-05-04
feat: DoiValidator

axthosarouris:
Copy DoiValidator from nva-fetch-doi to nva-commons, for usage in the nva-doi-trasformer.


axthosarouris:
> I created a task for this earlier: https://unit.atlassian.net/browse/NP-854

I need a validator for the nva-doi-transformer. We can improve the validation method when we address NP-854

PR #47, opened 2020-05-08
Add singleton collector

brinxmat:
This adds a SingletonCollector that provides 

- SingletonCollector::collect method that returns an object of type T where one and only one response is expected, or throws an error if zero or multiple objects are present
- SingletonCollector::collectOrElse that returns an object of type T where one and only one response is expected, or if no object is present returns the provided object, or throws an error if multiple objects are present

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,11 +6,16 @@ plugins {
     id 'jacoco'
     id 'checkstyle'
     id 'pmd'
+    id 'nebula.lint' version '16.8.0'
 }
 
 
 sourceCompatibility = JavaVersion.VERSION_11
 targetCompatibility = JavaVersion.VERSION_11
+
+gradleLint {
+    rules = ['unused-dependency']
+}
 
 wrapper {
     distributionType = Wrapper.DistributionType.ALL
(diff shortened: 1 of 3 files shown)


src/main/java/nva/commons/utils/SingletonCollector.java
```diff
@@ -0,0 +1,44 @@
+package nva.commons.utils;
+
+import java.util.List;
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+
+public class SingletonCollector {
```
axthosarouris, line 7: classes with only static methods should be final and with a private constructor.

build.gradle
```diff
@@ -50,6 +52,8 @@ dependencies {
     testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter', version: jupiterVersion
     testImplementation group: 'org.mockito', name: 'mockito-core', version: '3.3.3'
     testImplementation group: 'org.hamcrest', name: 'hamcrest', version: '2.2'
+    compileOnly group: 'org.apiguardian', name: 'apiguardian-api', version: '1.1.0'
```
janbj, line 55: What is this used for?

axthosarouris, line 55: this library is not necessary. remove?

PR #57, opened 2020-06-02
feat: isEmpty for strings

axthosarouris:


src/main/java/nva/commons/utils/StringUtils.java
```diff
@@ -35,4 +37,8 @@ public static String removeMultipleWhiteSpaces(String input) {
     public static String removeWhiteSpaces(String input) {
         return input.replaceAll(WHITESPACES, EMPTY_STRING);
     }
+
```
brinxmat, line 40: \s matches \t \n \r, there are cases where we want medial whitespaces, but not these or leading/trailing spaces…I implemented this in a recent PR, but I guess it belongs here

src/test/java/nva/commons/utils/StringUtilsTest.java
```diff
@@ -46,4 +46,35 @@ public void removeWhiteSpacesRemovesWhiteSpaces() {
         String output = StringUtils.removeWhiteSpaces(combinationOfSpaces);
         assertThat(output, is(equalTo(expected)));
     }
+
```
brinxmat, line 49: These could be ParametrizedTest and NullAndEmptySource

src/main/java/nva/commons/utils/StringUtils.java
axthosarouris: We can fix this, in another PR.

build.gradle
```diff
@@ -64,14 +64,14 @@ dependencies {
     implementation group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.9'
 
     testImplementation group: 'com.github.BIBSYSDEV', name: 'nva-testutils', version: '0.1.4'
-    testRuntimeOnly(group: 'org.junit.jupiter', name: 'junit-jupiter', version: jupiterVersion) {
+
+    testImplementation(group: 'org.junit.jupiter', name: 'junit-jupiter', version: jupiterVersion) {
```
axthosarouris, line 68: testRuntimeOnly is not sufficient, gradle complains that it does not find the annotations. Strangely enough gradle lint complains that the whole library is not necessary

PR #68, opened 2020-06-26
Add orElseThrow

brinxmat:
Added orElseThrow to tackle a common pattern that I see repeated in the code where we want to throw a specific exception.


src/main/java/nva/commons/utils/SingletonCollector.java
```diff
@@ -25,15 +28,32 @@ private SingletonCollector() {
         return Collectors.collectingAndThen(Collectors.toList(), SingletonCollector::get);
     }
 
-    private static <T> T get(List<T> list) {
+    public static <T> Collector<T, ?, T> collectOrElse(T alternative) {
```
axthosarouris, line 31: this was just moved.

PR #81, opened 2020-08-14
Np 1261 lambda authorizer for services

axthosarouris:
### Changes:

ApiGatewayHandler was split in the following way:

1) The RestRequestHandler (superclass)
2) The ApiGatewayHandler (subclass of RestRequestHandler)
3) The ServiceAuthorizerHandler (subclass of RestRequestHandler)

* The RestRequestHandler contains almost no new code.
* The code of the old ApiGatewayHandler was split between the RestRequestHandler and the new ApiGatewayHandler
* The RequestAuthorizer implements the RestRequestHandler and implements the logic of a
"request-based Lambda authorizer function " (see link below).
    * Example implementations of a RequestAuthorizer can be found in RequestAuthorizerTest and [here](https://raw.githubusercontent.com/BIBSYSDEV/nva-user-access-service/b35aa9abdc8dea7eff127c1c068416eb19b81f26/src/main/java/no/unit/nva/handlers/authorizer/LambdaAuthorizer.java)

### Details about the ServiceAuthorizerHandler.

1. The RequestAuthorizer is an abstract class that will be extended by each service authorizer of an endpoint.
2. The RequestAuthorizer has the following process-flow
    a) It  reads a Lambda ARN and an APi key from an HTTP request (the Lambda ARN is added by API Gateway)  
    b) it compares the client API key with the expected API key
    c) it returns a policy that allows the invocation of the specified lambda function.
3. The logic that provides the expected API key can be different from case to case and it is implemented by the handler that implements the RequestAuthorizer.
4. The logic that provides the necessary principal-id (i.e. who/what called the API) is also implementer by the handler that implements the RequestAuthorizer. 

 
### AWS Lambda Authorizers
https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html


src/main/java/nva/commons/handlers/ApiGatewayHandler.java
```diff
@@ -1,184 +1,60 @@
 package nva.commons.handlers;
 
-import static java.util.Objects.isNull;
+import static nva.commons.utils.JsonUtils.objectMapper;
```
axthosarouris, line 3: The old ApiGatewayHandler has been split into two classes now to reduce code duplication

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
```diff
@@ -1,184 +1,60 @@
...
- * @see <a href="https://github.com/awslabs/aws-serverless-java-container">aws-serverless-container</a> for
- *     alternative solutions.
- */
-public abstract class ApiGatewayHandler<I, O> implements RequestStreamHandler {
+public abstract class ApiGatewayHandler<I, O> extends RestRequestHandler<I, O> {
 
     public static final String ACCESS_CONTROL_ALLOW_ORIGIN = "Access-Control-Allow-Origin";
-    public static final String CONTENT_TYPE = "Content-Type";
-    public static final String APPLICATION_PROBLEM_JSON = "application/problem+json";
+    public static final String ALLOWED_ORIGIN_ENV = "ALLOWED_ORIGIN";
```
axthosarouris, line 31: moved from line 62

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
```diff
@@ -219,57 +95,39 @@ protected void setAdditionalHeadersSupplier(Supplier<Map<String, String>> additi
...
-     * @throws IOException when parsing fails.
-     */
-    protected I parseInput(String inputString) throws IOException {
-        return inputParser.getBodyElementFromJson(inputString, getIClass());
-    }
-
-    /**
-     * This is the message for the success case. Sends a JSON string containing the response that APIGateway will send
-     * to the user.
+     * Sends a message to ApiGateway and to the API-client, in case of failure caused by an ApiGatewayException
```
axthosarouris, line 98: moved from line 262

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
```diff
@@ -219,57 +95,39 @@ protected void setAdditionalHeadersSupplier(Supplier<Map<String, String>> additi
...
      */
-    protected void writeOutput(I input, O output)
-        throws IOException, GatewayResponseSerializingException {
-        try (BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(outputStream))) {
-            GatewayResponse<O> gatewayResponse = new GatewayResponse<>(output, getSuccessHeaders(),
-                getSuccessStatusCode(input, output));
-            String responseJson = objectMapper.writeValueAsString(gatewayResponse);
-            writer.write(responseJson);
+    @Override
+    protected void writeExpectedFailure(I input, ApiGatewayException exception, String requestId) throws IOException {
```
axthosarouris, line 107: line 270 in old version

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
```diff
@@ -219,57 +95,39 @@ protected void setAdditionalHeadersSupplier(Supplier<Map<String, String>> additi
...
+     * ApiGatewayException (unpredicted exception). This method can be overriden for richer status codes, but in the
+     * general case it should not be necessary.
      *
      * @param input     the input object of class I
      * @param exception the exception
      * @throws IOException when serializing fails
      */
-    protected void writeExpectedFailure(I input, ApiGatewayException exception, String requestId) throws IOException {
+    @Override
+    protected void writeUnexpectedFailure(I input, Exception exception, String requestId)
```
axthosarouris, line 125: line 315 in old version

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
```diff
@@ -304,23 +163,14 @@ protected void writeFailure(Exception exception, Integer statusCode, String requ
...
-    protected void writeUnexpectedFailure(I input, Exception exception, String requestId)
-        throws IOException {
-        try {
-            RuntimeException runtimeException =
-                new RuntimeException(MESSAGE_FOR_RUNTIME_EXCEPTIONS_HIDING_IMPLEMENTATION_DETAILS_TO_API_CLIENTS);
-            writeFailure(runtimeException, HttpStatus.SC_INTERNAL_SERVER_ERROR, requestId);
-        } catch (GatewayResponseSerializingException e) {
-            throw new ApiGatewayUncheckedException(e);
-        }
+    protected Map<String, String> getFailureHeaders() {
```
axthosarouris, line 170: line 226 in old version

src/main/java/nva/commons/handlers/ApiGatewayHandler.java
```diff
@@ -330,27 +180,26 @@ protected void writeUnexpectedFailure(I input, Exception exception, String reque
...
+    /**
+     * This is the message for the success case. Sends a JSON string containing the response that APIGateway will send
+     * to the user.
+     *
+     * @param input  the input object of class I
+     * @param output the output object of class O
+     * @throws IOException when serializing fails
+     */
+    @Override
+    protected void writeOutput(I input, O output)
```
axthosarouris, line 192: line 251 in old version

src/main/java/nva/commons/handlers/RestRequestHandler.java
```diff
@@ -0,0 +1,193 @@
+package nva.commons.handlers;
```
axthosarouris, line 1: There is virtuallly no new code in this class. All code pre-existed in ApiGatewayHandler

src/main/java/nva/commons/handlers/RestRequestHandler.java
```diff
@@ -0,0 +1,193 @@
...
+            init(output, context);
+            String inputString = IoUtils.streamToString(input);
+            inputObject = parseInput(inputString);
+
+            O response;
+            response = processInput(inputObject, inputString, context);
+
+            writeOutput(inputObject, response);
+        } catch (ApiGatewayException e) {
+            handleExpectedException(context, inputObject, e);
```
axthosarouris, line 72: the lines that were here were refactored to a method.

src/main/java/nva/commons/handlers/RestRequestHandler.java
```diff
@@ -0,0 +1,193 @@
...
+            handleTypeIdException(context, inputObject, e);
+        } catch (Exception e) {
+            handleUnexpectedException(context, inputObject, e);
+        }
+    }
+
+    protected void handleUnexpectedException(Context context, I inputObject, Exception e) throws IOException {
+        logger.error(e.getMessage());
+        logger.error(getStackTraceString(e));
+        writeUnexpectedFailure(inputObject, e, context.getAwsRequestId());
```
axthosarouris, line 83: This method is now abstract

src/main/java/nva/commons/handlers/RestRequestHandler.java
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    protected void handleTypeIdException(Context context, I inputObject, InvalidTypeIdException e) throws IOException {
+        logger.warn(e.getMessage());
+        logger.warn(getStackTraceString(e));
+        InvalidOrMissingTypeException apiGatewayInvalidTypeException = transformExceptionToApiGatewayException(e);
+        writeExpectedFailure(inputObject, apiGatewayInvalidTypeException, context.getAwsRequestId());
```
axthosarouris, line 90: This method is now abstract and implemented in ApiGatewayHandler and ServiceAuthorizerHandler

src/main/java/nva/commons/handlers/authentication/AuthPolicy.java
```diff
@@ -0,0 +1,92 @@
+package nva.commons.handlers.authentication;
```
axthosarouris, line 1: implements JSON objects like  following:

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": "execute-api:Invoke",
      "Effect": "Allow",
      "Resource": "arn:aws:execute-api:us-east-1:123456789012:ivdtdhp7b5/ESTestInvoke-stage/GET/"
    }
  ]
}
```

src/main/java/nva/commons/handlers/authentication/AuthorizerResponse.java
```diff
@@ -0,0 +1,73 @@
+package nva.commons.handlers.authentication;
```
axthosarouris, line 1: Response of the RequestAuthorizer

src/main/java/nva/commons/handlers/authentication/RequestAuthorizer.java
```diff
@@ -0,0 +1,166 @@
...
+import nva.commons.handlers.RequestInfo;
+import nva.commons.handlers.RestRequestHandler;
+import nva.commons.utils.Environment;
+import nva.commons.utils.JsonUtils;
+import nva.commons.utils.attempt.Failure;
+import org.apache.http.HttpHeaders;
+import org.apache.http.HttpStatus;
+import org.slf4j.LoggerFactory;
+
+public abstract class RequestAuthorizer extends RestRequestHandler<Void, AuthorizerResponse> {
```
axthosarouris, line 23: Implements the request-based authorizer as described [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html).

src/main/java/nva/commons/handlers/authentication/StatementElement.java
```diff
@@ -0,0 +1,102 @@
+package nva.commons.handlers.authentication;
```
axthosarouris, line 1: part of the AuthPolicy JSON object

src/test/java/nva/commons/handlers/ApiGatewayHandlerTest.java
```diff
@@ -126,11 +128,26 @@ public void apiGatewayHandlerAllowsResponseHeadersDependedOnInput() throws IOExc
...
         ByteArrayOutputStream outputStream = outputStream();
         handler.handleRequest(input, outputStream, context);
-        String outputString = outputStream.toString(StandardCharsets.UTF_8);
-        GatewayResponse<String> response = objectMapper.readValue(outputString, GatewayResponse.class);
+        GatewayResponse<String> response = GatewayResponse.fromOutputStream(outputStream);
         assertTrue(response.getHeaders().containsKey(HttpHeaders.WARNING));
     }
 
+    @Test
+    public void handleRequestDoesNotThrowExceptionOnUnknownFieldsInRequestInfo() throws IOException {
```
axthosarouris, line 136: Double-checking that unknown fields in RequestInfo actually works. Tests in RequestInfo exist as well.

src/main/java/nva/commons/handlers/authentication/RequestAuthorizer.java
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * <p>Example methodARN:
+     * arn:aws:execute-api:eu-west-1:884807050265:2lcqynkwke/Prod/GET/service/users/orestis@unit.no Example output:
+     * arn:aws:execute-api:eu-west-1:884807050265:2lcqynkwke/Prod\/*\/* <br/> Another possible output is: "*"
+     *
+     * @param methodArn the method ARN as provided by the API gateway
+     * @return a resource for the policy
+     */
+    protected String formatResource(String methodArn) {
+        String[] resource2 = methodArn.split(PATH_DELIMITER);
```
brinxmat, line 63: resource2 is a list of resource path components? I guess you thought resource2X, but I think resourcePathComponents maybe makes more sense as I was looking for the original resource in the method.

brinxmat: Although, in the given case, finding the second index of "/" using https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#indexOf-java.lang.String-int- and adding the wild-carded path might be better as it does more like what you are actually doing (in my mind)

src/main/java/nva/commons/handlers/authentication/RequestAuthorizer.java
```diff
@@ -0,0 +1,166 @@
...
+     * allowed to. It can contain wildcards.
+     *
+     * <p>Example methodARN:
+     * arn:aws:execute-api:eu-west-1:884807050265:2lcqynkwke/Prod/GET/service/users/orestis@unit.no Example output:
+     * arn:aws:execute-api:eu-west-1:884807050265:2lcqynkwke/Prod\/*\/* <br/> Another possible output is: "*"
+     *
+     * @param methodArn the method ARN as provided by the API gateway
+     * @return a resource for the policy
+     */
+    protected String formatResource(String methodArn) {
```
brinxmat, line 62: The method name here intends to remove the user and resource specifics of the ARN, maybe it should be named to reflect this?

src/main/java/nva/commons/handlers/authentication/RequestAuthorizer.java
```diff
@@ -0,0 +1,166 @@
...
+        String[] resource2 = methodArn.split(PATH_DELIMITER);
+        String apiGateway = resource2[API_GATEWAY_IDENTIFIER_INDEX];
+        String stage = resource2[STAGE_INDEX];
+        String method = WILDCARD;
```
brinxmat, line 66: Inline vars

src/main/java/nva/commons/handlers/authentication/RequestAuthorizer.java
```diff
@@ -0,0 +1,166 @@
...
+            .build();
+        return AuthPolicy.newBuilder().withStatement(Collections.singletonList(statement)).build();
+    }
+
+    protected abstract String principalId() throws ForbiddenException;
+
+    protected abstract String fetchSecret() throws ForbiddenException;
+
+    protected void secretCheck(RequestInfo requestInfo) throws ForbiddenException {
+        if (requestInfo.getHeaders().containsKey(HttpHeaders.AUTHORIZATION)) {
```
brinxmat, line 96: Can this be refactored so that clientSecret is Optional.ofNullable…orElseThrow? Removing the need for the outer if and the null-check in the second if?

brinxmat: Assuming that clientSecret. being "null" is in fact an error, this is a more specific error than Forbidden

src/main/java/nva/commons/handlers/authentication/RequestAuthorizer.java
axthosarouris: I thought to be explicit on what the wildcard is for.

src/main/java/nva/commons/handlers/authentication/RequestAuthorizer.java
axthosarouris: Actually no. The default operation is to remove the HttpMethod and the path of the called function. This is due to the fact that ApiGateway caches the policy and does not run the authorizer for subsequent calls for 300 secs (configurable). This means that if I call GET /user/orestis, and I keep the method and the path then I can only GET /user/orestis. That is why I use the wildcards. Once the secret matching is done, the service gets permission to perform any method on any path of that ApiGateway instance. 

This method is overridable, so that if for some service the requirements for the access policy are different, then the custom handler will be able to do whatever they want with the resource policy.

As a reference when I say resource I mean the resource in the followgin:

```
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Action": "execute-api:Invoke",
      "Effect": "Allow",
      "Resource": "arn:aws:execute-api:us-east-1:123456789012:ivdtdhp7b5/ESTestInvoke-stage/GET/"
    }
  ]
}
```

src/main/java/nva/commons/handlers/authentication/RequestAuthorizer.java
axthosarouris: The specific error is logged. The client does not need to know anything else except for the fact they do not have access. Any more information about why the client is not allowed to have access is irrelevant and a potential threat

PR #105, opened 2020-09-30
NP-1496: Create handler that assumes role 

axthosarouris:


build.gradle
```diff
@@ -66,7 +68,7 @@ dependencies {
     implementation group: 'org.apache.httpcomponents', name: 'httpcore', version: '4.4.11'
     implementation group: 'org.apache.httpcomponents', name: 'httpclient', version: '4.5.9'
 
-    testImplementation group: 'com.github.BIBSYSDEV', name: 'nva-testutils', version: '0.1.13'
+    testImplementation group: 'com.github.BIBSYSDEV', name: 'nva-testutils', version: 'db959439b7d354b'
```
norrs, line 71: Remember to update.

src/main/java/nva/commons/handlers/AuthorizedHandler.java
```diff
@@ -0,0 +1,59 @@
...
+
+    protected abstract O processInput(I input,
+                                      RequestInfo requestInfo,
+                                      STSAssumeRoleSessionCredentialsProvider credentialsProvider,
+                                      Context context) throws ApiGatewayException;
+
+    protected abstract List<Tag> sessionTags(RequestInfo requestInfo) throws ApiGatewayException;
+
+    protected String assumedRoleArn() {
+        return environment.readEnv(ASSUMED_ROLE_ARN_ENV_VAR);
```
norrs, line 44: Since this is read from environment, isn't this environment set by the instance it is running under the hood in AWS Lambda? 

I noticed there were defined `AssumedRole` and `AdminRole` during your demostration . Will we have roles according to defined privileged access in our applications/services? Ie `User`, `Curator`, `Editor` , `Administrator`, `Application Administrator` ? 

And since this then comes from the environment variable, wouldn't we need to deploy the same function multiple times for each role with associated policies? ie because an administrator or application administrator is allowed to have different access policies associated with more access?

norrs: I guess the different `roles` can have different policies?

axthosarouris: Just ignore AdminRole. It was a super role that I used to be able to demonstrate the functionality of the new handler without having to refactor all the handlers. So it is not relevant. 

It is not necessary to have roles according to the application roles, i.e., User, Curator, ..etc. Depending on how you define the conditions on the role, one role could dynamically according to the tags that you provide to the role. 

The AssumedRole is basically saying this: Give me a tag and i will tell you what you have access to. This is resolved in runtime, since the tag is supplied in runtime.

The flow is the following
DeployTime:
1. Create the following:
    * LambdaRole with no access to DynamoDB
    * AssumedRole with a Condition that will be evaluated in runtime.
2. Provide the AssumedRole.ARN to the Lambda function as env variable

Runtime
1. Get the feideId or other info from the bearer token
2. Assume the role with the generated tags based on step (1)  The condition evaluation happesns now given the tags that we provided.
3. "Log in" the database with the assumed role.

Links:
https://docs.aws.amazon.com/STS/latest/APIReference/API_AssumeRole.html
https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/specifying-conditions.html
https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_request.html

src/main/java/nva/commons/handlers/RequestInfo.java
```diff
@@ -167,6 +167,11 @@ public JsonNode getRequestContext() {
         return requestContext;
     }
 
+    /**
```
axthosarouris, line 170: checkstyle complains.

norrs (approved):
Remember to update nva-testutils.

PR #110, opened 2020-10-14
NP-1668:exception message in one line util method

axthosarouris:
Moved existing code from RestRequestHandler to StringUtils for reusing it in EventHandler.


norrs (changes requested):
Maybe put it as `StacktraceUtils` ?  
Its verify specific to `Stacktrace /Exceptionutils` and not `String`.

brinxmat (changes requested):
StackTraceUtils +1

brinxmat (approved):

PR #111, opened 2020-10-19
Utility for string is neither null or empty

norrs:


src/main/java/nva/commons/utils/StringUtils.java
```diff
@@ -51,4 +52,22 @@ public static boolean isEmpty(String input) {
...
         return str.replaceAll("\\s", " ");
     }
+
+    /**
+     * Checks if string is neither null or empty.
+     * @param string string to check
+     * @return <code>true</code> if string is neither null or empty.
+     */
+    public static boolean isNotNullOrEmpty(String string) {
+        return nonNull(string) && !string.isEmpty();
```
axthosarouris, line 62: why not simply call !(isNullOrEmpty)? better to reuse code and also to show the intention?

axthosarouris (changes requested):

axthosarouris (approved):

PR #126, opened 2020-11-12
isEmpty is not the same as isBlank!

norrs:


src/main/java/nva/commons/utils/StringUtils.java
```diff
@@ -38,10 +38,14 @@ public static String removeWhiteSpaces(String input) {
         return input.replaceAll(WHITESPACES, EMPTY_STRING);
     }
 
-    public static boolean isEmpty(String input) {
+    public static boolean isBlank(String input) {
```
axthosarouris, line 41: this is the same as the isNullOrBlank

norrs: Nice catch!

src/main/java/nva/commons/utils/StringUtils.java
```diff
@@ -38,8 +38,26 @@ public static String removeWhiteSpaces(String input) {
         return input.replaceAll(WHITESPACES, EMPTY_STRING);
     }
 
+    /**
+     * Checks if string input is blank.
+     * @param input input string
+     * @return <code>true</code> if blank.
+     * @see #isNullOrBlank(String) 
+     */
+    @Deprecated(forRemoval = true)
```
norrs, line 47: @axthosarouris : Are we ok with this deprecation? 

I think we should deprecate at least one of the versions.. 
Maybe keep the short version instead and deprecate the other with longer names and add description in javadoc that is null safe?

src/main/java/nva/commons/utils/StringUtils.java
norrs: I deprecated the isNullOrBlank version!

axthosarouris (dismissed):

src/main/java/nva/commons/utils/StringUtils.java
```diff
@@ -38,10 +38,25 @@ public static String removeWhiteSpaces(String input) {
...
     }
 
+    /**
+     * Checks if string input is empty or null.
+     *
+     * @param input input string
+     * @return <code>true</code> if empty.
+     */
+    public static boolean isEmpty(String input) {
+        return isNull(input) || input.isEmpty();
```
norrs, line 57: Before this method checked `input.isBlank` , so this is a breaking change!

norrs: Also introduced `isBlank` that you can use if bumping library and tests fail!

norrs:
FYI @svenngjeto @janbj @brinxmat @Eiriknil

src/test/java/nva/commons/utils/StringUtilsTest.java
```diff
@@ -56,41 +59,85 @@ public void stringUtilsHasAMethodForCheckingIfAStringIsEmpty() {
...
-    @DisplayName("isNotNullOrEmpty should return true where string is neither null or empty.")
+    @DisplayName("isNotNullOrEmpty should return true where string is neither null or empty")
+    @Deprecated(forRemoval = true)
     public void stringUtilIsNotNullOrEmpty(String input) {
         assertThat(StringUtils.isNotNullOrEmpty(input), is(equalTo(true)));
     }
 
-    @ParameterizedTest
+    @ParameterizedTest(name = PARAMETERIZED_TEST_NAME_FORMAT)
+    @ValueSource(strings = {" ", "\t", "\n", "a"})
```
brinxmat, line 106: Need to include everything that is blank or null when we state we are testing for it. Either multiple tests (one with @NullSource, or as follows).
`
Stream<String> blankStrings() {
    Stream.of("", "  ",  "\t", "\n", "\r\n", null); // non-exhaustive list…!
}

@ParameterizedTest(name = "…")
@MethodSource("blankStrings")
`

brinxmat (changes requested):

src/test/java/nva/commons/utils/StringUtilsTest.java
```diff
@@ -56,41 +59,85 @@ public void stringUtilsHasAMethodForCheckingIfAStringIsEmpty() {
         boolean result = StringUtils.isEmpty(null);
     }
 
-    @Test
-    @DisplayName("isEmpty should return false for a non empty string")
-    public void isEmptyReturnsFalseForANonEmptyString() {
-        String nonEmptyString = "abc";
-        assertThat(StringUtils.isEmpty(nonEmptyString), is(equalTo(false)));
+    @ParameterizedTest(name = PARAMETERIZED_TEST_NAME_FORMAT)
```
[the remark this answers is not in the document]
norrs: I disagree, we should not drop **@DisplayName** . 

Because name of the @ParameterizedTest can simply use the static template for all Parameterized Tests, then description of test name is done under DisplayName which is the same for @ParameterizedTest and @Test .

src/test/java/nva/commons/utils/StringUtilsTest.java
norrs: Added.

axthosarouris (approved):

norrs:
@axthosarouris @brinxmat @janbj Any input regarding this: https://github.com/google/guava/wiki/UsingAndAvoidingNullExplained#convenience-methods ?

src/test/java/nva/commons/utils/StringUtilsTest.java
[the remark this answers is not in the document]
brinxmat: Fair enough.I was previously a fan of DisplayName, but I see that maintenance has not been followed up in most cases, we will have to remember to follow up naming conventions in future.

brinxmat (approved):

src/test/java/nva/commons/utils/StringUtilsTest.java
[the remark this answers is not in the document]
norrs: @brinxmat FYI I include the `@DisplayName` in the `@ParameterizedTest(name= ..`  . So this is just enabling `@DisplayName` to be shown for `@ParameterizedTest` as well as long as they referer to the TEST NAME FORMAT const.

norrs:
We should consider adding:

```
emptyToNull(String)
isNullOrEmpty(String)
nullToEmpty(String)
```

PR #132, opened 2020-12-09
IoUtils for reading from resources taking path as string

janbj:


src/main/java/nva/commons/utils/IoUtils.java
```diff
@@ -18,6 +18,8 @@
 public final class IoUtils {
 
     public static final String LINE_SEPARATOR = System.lineSeparator();
+    public static final String REGEX = "\\\\";
```
axthosarouris, line 21: WIN_PATH_SEPARATOR_REGEX

src/main/java/nva/commons/utils/IoUtils.java
```diff
@@ -18,6 +18,8 @@
...
+    public static final String REPLACEMENT = "/";
```
axthosarouris, line 22: PATH_SEPARATOR_FOR_RESOURCES

src/main/java/nva/commons/utils/IoUtils.java
```diff
@@ -29,10 +31,22 @@ private IoUtils() {
      * @param path the Path to the resource.
      * @return an InputStream with the data.
      */
+    @Deprecated
     public static InputStream inputStreamFromResources(Path path) {
+        String pathString = path.toString().replaceAll(REGEX, REPLACEMENT);
```
axthosarouris, line 36: replaceWinPathSeparatorsWithUniversalPathSeparators

axthosarouris (changes requested):
Few comments about naming, other than that very elegant :)

axthosarouris (approved):

PR #143, opened 2021-01-14
Remove obsolete parameter-names package

axthosarouris:


build.gradle
```diff
@@ -164,6 +161,11 @@ subprojects{
         testImplementation group: 'org.hamcrest', name: 'hamcrest', version: '2.2'
         testImplementation group: 'com.github.npathai', name: 'hamcrest-optional', version: '2.0.0'
         testImplementation group: 'com.github.BIBSYSDEV', name: 'nva-testutils', version: '0.1.14'
+
+
+        compileOnly (group: 'org.apiguardian', name: 'apiguardian-api', version: '1.1.1'){
+            because "gradle complains"
```
brinxmat, line 167: Why does Gradle complain? IIRC because some module depends on apiguardian, possibly better to annotate this so we can follow it up more precisely when the issue is fixed. It was JUnit, IIRC, but maybe this is fixed in the latest release.

brinxmat (changes requested):

janbj (approved):

brinxmat (approved):

PR #155, opened 2021-01-29
Np 1679 http headers

axthosarouris:


build.gradle
```diff
@@ -189,6 +189,7 @@ subprojects {
         pkg {
             repo = 'nva-commons'
             name = project.name
+            userOrg = 'unit-nva'
```
axthosarouris, line 192: This comes from develop, I think. It is part of Jan's PR.

brinxmat (approved):

PR #159, opened 2021-02-05
Created commons dynamodb module with ItemMapper

janbj:


dynamodb/build.gradle
```diff
@@ -0,0 +1,30 @@
+dependencies {
+    api project (":core")
+
+    api(group: 'com.amazonaws', name: 'aws-java-sdk-dynamodb', version: project.ext.awsSdkVersion) {
+
```
janbj, line 5: I will take a check for exclusions here

brinxmat (approved):

dynamodb/src/main/java/nva/commons/dynamodb/ItemMapper.java
```diff
@@ -0,0 +1,151 @@
...
+            Map<String, software.amazon.awssdk.services.dynamodb.model.AttributeValue> attributeValueMapV2)
+            throws JsonProcessingException {
+        Map<String, AttributeValue> attributeValueMap = fromV2toV1(attributeValueMapV2);
+        return toJsonNode(attributeValueMap);
+    }
+
+    public static JsonNode toJsonNode(Map<String, AttributeValue> attributeValueMap) {
+        var object = toObject(attributeValueMap);
+        Item item = Item.fromMap(object);
+        var jsonNode = objectMapper.convertValue(item.toJSON(), JsonNode.class);
```
axthosarouris, line 54: This converts the String produced by the "toJson()" to a StringNode. The result here is a StringnNode. The final result will be 
```
" {\"id\":\"value\"}" instead of a real json string.
```

dynamodb/src/test/java/nva/commons/dynamodb/ItemMapperTest.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @Test
+    public void toJsonNodeReturnsJsonOnAttributeValueMap() throws IOException {
+        var file = new File(ATTRIBUTE_VALUE_MAP_JSON);
+        var map = objectMapper.readValue(file, HashMap.class);
+        var item = Item.fromMap(map);
+        var attributeValueMap = ItemUtils.toAttributeValues(item);
+        var json = ItemMapper.toJson(attributeValueMap);
+
+        assertThat(json, is(notNullValue()));
```
axthosarouris, line 28: The json string is not a Json string. Maybe we should create a small Dynamo Json file and the expected Json and compare them. Now we do not know if we are losing any information. 
Better would be if we used DynamoDb local to insert an object and read it again using our (de)/serializer. Then we would know it works.

axthosarouris (changes requested):
Requested changes: the above comments.

General comment: We need a local database for our tests. Without it I do not think we can migrate to sdk2. It seems that there are two alternatives (listed below).  Do we consider the usage of a new local dynamodb to be part of this task?


* https://hub.docker.com/r/amazon/dynamodb-local/
* https://github.com/aws/aws-sdk-java-v2/blob/93269d4c0416d0f72e086774265847d6af0d54ec/services-custom/dynamodb-enhanced/src/test/java/software/amazon/awssdk/extensions/dynamodb/mappingclient/functionaltests/LocalDynamoDb.java

axthosarouris:
Do we have both develop and main? Should we protect develop ?

dynamodb/src/main/java/nva/commons/dynamodb/ItemMapper.java
janbj: Hm,  had a reason for this originally as a part of trying to read to a v2 Map<String,AttributeValue> but you are of course right.

brinxmat (dismissed):

build.gradle
```diff
@@ -17,7 +17,7 @@ wrapper {
 
 allprojects {
     group = 'no.unit.nva'
-    version = '1.0.0'
+    version = '1.0.2'
```
axthosarouris, line 20: Why to merge in develop? Version number is 1.0.3

axthosarouris (approved):

axthosarouris (changes requested):
wrong branch?

axthosarouris (approved):

PR #165, opened 2021-02-23
identifier from URI

axthosarouris:


identifiers/src/main/java/no/unit/nva/identifiers/SortableIdentifier.java
```diff
@@ -32,6 +34,16 @@ public static SortableIdentifier next() {
         return new SortableIdentifier(newIdentifierString());
     }
 
+    public static SortableIdentifier fromUri(URI uri) {
+        try {
+            String path = uri.getPath();
+            String[] pathElements = path.split(PATH_DELIMITER);
```
brinxmat, line 40: Probably just

            return new SortableIdentifier(path.substring(path.lastIndexOf(PATH_DELIMITER) + 1));

Which avoids firing up the regex and the jiggery pokery of arrays

brinxmat (changes requested):

brinxmat (approved):

PR #166, opened 2021-02-24
Np 2220 update GitHub actions definition

mattiahj:


build.gradle
```diff
@@ -5,33 +5,35 @@ plugins {
...
     apply plugin: 'jacoco'
     apply plugin: 'base'
     apply plugin: 'checkstyle'
     apply plugin: 'pmd'
-    apply plugin: 'com.jfrog.bintray'
 
     sourceCompatibility = JavaVersion.VERSION_11
     targetCompatibility = JavaVersion.VERSION_11
 
+    // Needed because our javadoc contains a bunch of errors. TODO: Fix these errors.
```
mattiahj, line 30: Normally I might have done this straight away, but I remembered you guys want as small pull requests as possible.

build.gradle
```diff
@@ -5,33 +5,35 @@ plugins {
...
+    tasks.withType(Javadoc) {
+        options.addStringOption('Xdoclint:none', '-quiet')
+    }
+
     repositories {
-        jcenter()
-        maven {   url  "https://dl.bintray.com/unit-nva/nva-commons"}
+        mavenCentral()
         maven { url 'https://jitpack.io' }
```
mattiahj, line 37: Can we remove this? We seem to depend on some of our own artifacts we've published there. Perhaps we can do away with it after having made the switch to Maven Central?

axthosarouris: When we put things in maven central, then we can move it. We should update first all projects though, otherwise some of them will not be built

axthosarouris (approved):

build.gradle
brinxmat: I guess that the PMD-config needs updating too

brinxmat (approved):

brinxmat (approved):

PR #171, opened 2021-03-02
S3 client

axthosarouris:
Simple S3 client class for getting files. It will be used to read DynamoDB exports to S3.


s3/src/test/java/no/unit/nva/s3/S3DriverTest.java
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    @BeforeEach
+    public void init() throws IOException {
+        s3Client = mockS3Client();
+        s3Driver = new S3Driver(s3Client, SAMPLE_BUCKET);
+    }
+
+    @Test
+    @Tag("RemoteTest")
```
axthosarouris, line 86: "RemoteTests" are being ignored by now. They are there only to be run manually when necessary

brinxmat (approved):

janbj (approved):

svenngjeto (approved):

mattiahj (approved):

PR #182, opened 2021-03-16
missing query parameter throws bad-request exception

axthosarouris:


torarnet (approved):
Not sure if it is needed to test the BadRequestException itself since it is so trivial

brinxmat (approved):

PR #187, opened 2021-03-17
enable publishing to local repository

axthosarouris:
Enable publishing to local repository using `gradle publishToMavenLocal`

This means that the library will be available for usage locally by including the repository `mavenLocal()`

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ wrapper {
 
 allprojects {
     group = 'com.github.bibsysdev'
-    version = '1.5.0'
+    version = '1.5.1'
 
     apply plugin: 'java-library'
     apply plugin: 'jacoco'
--- a/build.gradle
+++ b/build.gradle
@@ -180,7 +180,16 @@ subprojects {
     publishing {
         publications {
             mavenJava(MavenPublication) {
+                artifactId = project.name
                 from components.java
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -210,11 +219,29 @@ subprojects {
     }
 
     signing {
-        useInMemoryPgpKeys(findProperty("signingKey"), findProperty("signingPassword"))
-        sign publishing.publications.mavenJava
+        if (isMainBranch()) {
```
axthosarouris, line 222: maybe not necessary to sign versions that are not going to be published in maven central.

build.gradle
```diff
@@ -210,11 +219,29 @@ subprojects {
...
+    def proc = "git rev-parse --abbrev-ref HEAD".execute()
+    proc.in.eachLine { line -> branch = line }
+    proc.err.eachLine { line -> println line }
+    proc.waitFor()
+    branch
+}
+
+boolean isMainBranch() {
+    def branch = gitBranch()
+    return ((branch == "master" || branch == "main"))
```
brinxmat, line 241: We never publish from develop, right?

brinxmat (approved):

build.gradle
axthosarouris: never

PR #188, opened 2021-03-18
Failure throws original unchecked exception

axthosarouris:
orElseThrow  in `attempt(()->someAction()).orElseThrow()` catches the excpection in thrown by someAction and throws a RuntimeException with cause the caught Exception. This is inevitable for checked exceptions , but for unchecked exceptions it is not necessary and it creates noise as one has to go and obscures the real failure.

This fix throws the original exception directly when it is a Runtime exception.

### DIFF
--- a/core/src/main/java/nva/commons/core/attempt/Failure.java
+++ b/core/src/main/java/nva/commons/core/attempt/Failure.java
@@ -44,6 +44,11 @@ public class Failure<T> extends Try<T> {
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


core/src/main/java/nva/commons/core/attempt/Failure.java
```diff
@@ -43,6 +43,11 @@ public Exception getException() {
         return new Failure<S>(exception);
     }
 
+    @Override
```
axthosarouris, line 46: automatic code re-arranging

brinxmat (approved):

PR #190, opened 2021-03-23
S3Driver with credentials from environment

axthosarouris:
Will be used for the Cristin import


brinxmat (approved):

torarnet (approved):
When working with another project i discovered that explicitly defining credentials was not always needed if it was a service living in AWS if it had the necessary role and policy attached to it. It might not be relevant in this case but just want to mention it

PR #191, opened 2021-03-23
Parallel mapper

axthosarouris:
This mapper was made having HTTP requests parallelization in mind.
`stream().parallel()` will create only as many threads as there are logical cores in the system and even that is not guaranteed. In case of lambdas there is no benefit of the above call, because in most cases lambdas see only one core.  

In case of parallelizing many HTTP calls the above approach is  not enough and using a lot of threads speeds up the process a lot. This has been previously tested (with @brinxmat ) although the test is not included in this PR. The speedup  is due to the fact that HTTP calls are very slow compared to any other operation happening in the (virtual) computer and while waiting for a response the processor can initiate many other HTTP calls.

### DIFF
--- a/core/src/main/java/nva/commons/core/ParallelMapper.java
+++ b/core/src/main/java/nva/commons/core/ParallelMapper.java
@@ -1,0 +1,93 @@
+package nva.commons.core;
+
+import static java.lang.Math.min;
+import static nva.commons.core.attempt.Try.attempt;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.function.Function;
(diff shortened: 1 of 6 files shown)


core/src/main/java/nva/commons/core/ParallelMapper.java
```diff
@@ -0,0 +1,93 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import nva.commons.core.attempt.Try;
+
+@SuppressWarnings("PMD.DoNotUseThreads")
```
brinxmat, line 17: Generally sound advice, sometimes unavoidable :D

brinxmat (approved):

brinxmat (approved):

svenngjeto (approved):

PR #194, opened 2021-03-26
S3 driver compresses data before stroring them

axthosarouris:


s3/src/main/java/no/unit/nva/s3/S3Driver.java
```diff
@@ -75,6 +78,20 @@ public void insertFile(Path fullPath, String content) {
         client.putObject(putObjectRequest, createRequestBody(content));
     }
 
+    public void insertFile(Path fullPath, InputStream content) throws IOException {
+        PutObjectRequest putObjectRequest = newPutObjectRequest(fullPath);
+        client.putObject(putObjectRequest, createRequestBody(content));
```
mattiahj, line 83: `client.putObject(newPutObjectRequest(fullPath), createRequestBody(content));` might suffice

s3/src/main/java/no/unit/nva/s3/S3Driver.java
```diff
@@ -120,6 +137,15 @@ private static void verifyThatRequiredEnvVariablesAreInPlace() {
         environment.readEnv(AWS_SECRET_ACCESS_KEY_ENV_VARIABLE_NAME);
     }
 
+    private RequestBody createRequestBody(InputStream compressedContent) throws IOException {
+        var bytes = IoUtils.inputStreamToBytes(compressedContent);
```
mattiahj, line 141: its subjective style stuff, but if the variable is only used once it can probably be inlined

mattiahj (dismissed):

s3/src/main/java/no/unit/nva/s3/S3Driver.java
axthosarouris: fixed

mattiahj (approved):

PR #205, opened 2021-04-09
Rename s3Driver methods for readability

axthosarouris:


s3/src/main/java/no/unit/nva/s3/S3Driver.java
```diff
@@ -113,7 +113,7 @@ public void insertAndCompressFiles(List<String> content) throws IOException {
         return resultBuffer;
     }
 
-    public Optional<String> getFile(Path file) {
+    public Optional<String> getUncompressedFile(Path file) {
```
axthosarouris, line 116: getFile renamed to getUncompressedFile

s3/src/main/java/no/unit/nva/s3/S3Driver.java
```diff
@@ -126,6 +126,14 @@ public String getCompressedFile(Path file) throws IOException {
         }
     }
 
+    public String getFile(String filename) {
```
axthosarouris, line 129: readFileContent renamed to getFile

brinxmat (approved):

PR #218, opened 2021-05-04
s3Driver allows writing in folders

axthosarouris:


svenngjeto (approved):

s3/src/test/java/no/unit/nva/s3/S3DriverTest.java
```diff
@@ -180,6 +184,23 @@ public void insertAndCompressFilesFilesStoresAllFilesCompressedInGzipStream() th
         assertThat(actualContent, is(equalTo(input)));
     }
 
+    @ParameterizedTest(name = "insertAndCompressFilesStoresAllFilesCompressedInGzipStreamInSpecifiedFolder")
+    @ValueSource(strings = {EMPTY_STRING, ROOT})
```
torarnet, line 188: Im not so experienced in using ParameterizedTest but how are you using 2 strings, `EMPTY_STRING, ROOT` but only one `String pathPrefix`

s3/src/main/java/no/unit/nva/s3/S3Driver.java
```diff
@@ -151,8 +160,18 @@ private static void verifyThatRequiredEnvVariablesAreInPlace() {
...
 
-    private Path filenameForZippedFile() {
-        return Path.of(UUID.randomUUID().toString() + GZIP_ENDING);
+    private Path filenameForZippedFile(Path s3Folder) {
+        String folderPath = processPath(s3Folder);
+        return Path.of(folderPath, UUID.randomUUID().toString() + GZIP_ENDING);
+    }
+
+    private String processPath(Path s3Folder) {
+        String unixPath = s3Folder.toString()
```
torarnet, line 169: Unix is AWS style, right? It should not be a problem when developing on Windows machines?

torarnet (approved):

s3/src/test/java/no/unit/nva/s3/S3DriverTest.java
axthosarouris: These are the values the parameter will take successively

s3/src/main/java/no/unit/nva/s3/S3Driver.java
axthosarouris: There is no local implementation of S3

PR #221, opened 2021-05-19
EventHandler handleError becomes overridable

axthosarouris:


eventhandlers/src/main/java/no/unit/nva/events/handlers/EventHandler.java
```diff
@@ -38,15 +38,16 @@ protected EventHandler(Class iclass) {
...
-            String inputString = IoUtils.streamToString(inputStream);
+            inputString = IoUtils.streamToString(inputStream);
             logger.trace(HANDLER_INPUT + inputString);
             AwsEventBridgeEvent<InputType> input = parseEvent(inputString);
             OutputType output = processInput(input.getDetail(), input, context);
 
             writeOutput(outputStream, output);
         } catch (Exception e) {
-            handleError(e);
+            handleError(e, inputString);
```
brinxmat, line 50: handleError doesn't actually use inputString

brinxmat (changes requested):

eventhandlers/src/main/java/no/unit/nva/events/handlers/EventHandler.java
brinxmat: It can also be null

axthosarouris: No, but I want to give the ability when we override the method to be able to use the inputString.

brinxmat (approved):

PR #223, opened 2021-06-03
NP-2366 Language code mapper converts ISO to URI

Eirik-Eide-Saur:
Implemented the features described in task NP-2366

### DIFF
--- a/core/src/test/java/nva/commons/core/language/LanguageMapperTest.java
+++ b/core/src/test/java/nva/commons/core/language/LanguageMapperTest.java
@@ -1,0 +1,78 @@
+package nva.commons.core.language;
+
+import static nva.commons.core.language.LanguageMapper.ERROR_MESSAGE_MISSING_RESOURCE_EXCEPTION;
+import static nva.commons.core.language.LanguageMapper.LEXVO_URI_PREFIX;
+import static nva.commons.core.language.LanguageMapper.LEXVO_URI_UNDEFINED;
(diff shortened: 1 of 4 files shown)


core/src/test/java/nva/commons/core/language/LanguageMapperTest.java
```diff
@@ -0,0 +1,78 @@
...
+import nva.commons.logutils.LogUtils;
+import nva.commons.logutils.TestAppender;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.NullAndEmptySource;
+import org.junit.jupiter.params.provider.ValueSource;
+
+public class LanguageMapperTest {
+
+    public static final String ISO3_LANG_CODE = "eng";
```
brinxmat, line 21: Call the codes what they (purport) to be eg. ISO_639_3_LANGUAGE_CODE and ISO_639_1_LANGUAGE_CODE. 

I guess that we will be surprised by ISO 639-2/B variant codes such as "arm" and "ger", so maybe worth testing if these produce expected output see https://gist.github.com/brinxmat/51823e80044462819e8e35b9b0725a53. I guess this is extra functionality, so create a task for this as the codes are unfortunately in common use in the bibliographic domain.

brinxmat (changes requested):

brinxmat (approved):

PR #227, opened 2021-06-09
Unix path serialized as string and deserialized from String

axthosarouris:


s3/src/test/java/no/unit/nva/s3/UnixPathTest.java
```diff
@@ -144,4 +150,50 @@ public void getFilenameReturnsTheLastElementOfaUnixPath(String inputPath, String
...
+    public void objectMapperSerializesUnixPathAsString() throws JsonProcessingException {
+        String unixPath = "/some/folder";
+
+        ClassWithUnixPath classWithUnixPath = new ClassWithUnixPath();
+        classWithUnixPath.setField(UnixPath.of(unixPath));
+        String jsonString = objectMapper.writeValueAsString(classWithUnixPath);
+        JsonNode actualJson = objectMapperNoEmpty.readTree(jsonString);
+
+        ObjectNode expectedJson = objectMapperNoEmpty.createObjectNode();
+        expectedJson.put(ClassWithUnixPath.fieldName(), unixPath);
```
axthosarouris, line 164: Not sure if I should write `String.format(" {\"field\" : \"%s\"}` for the expected Json just to be more clear of what is the expectation.

svenngjeto (approved):

s3/src/test/java/no/unit/nva/s3/UnixPathTest.java
brinxmat: Alternatively, put the unixpath into the ClassWithUnixPath constructor and simply look at the values of the json object to assert that there is a matching string, rather than an entire object.

brinxmat (approved):

PR #228, opened 2021-06-10
UnixPath: manipulate root

axthosarouris:
When manipulating URIs and S3 bucket paths we need to add or remove the root. Added this functionality in the UnixPath to avoid code replication.

### DIFF
--- a/s3/src/main/java/no/unit/nva/s3/UnixPath.java
+++ b/s3/src/main/java/no/unit/nva/s3/UnixPath.java
@@ -18,7 +18,7 @@ public final class UnixPath {
 
     public static final UnixPath EMPTY_PATH = new UnixPath(Collections.emptyList());
     private static final String PATH_DELIMITER = "/";
-    private static final String ROOT = "/";
+    public static final String ROOT = "/";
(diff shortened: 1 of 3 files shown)


s3/src/main/java/no/unit/nva/s3/UnixPath.java
```diff
@@ -74,43 +71,75 @@ public boolean equals(Object o) {
...
-    }
-
+    @JsonValue
     @Override
     public String toString() {
         if (pathIsEmpty(path)) {
             return EMPTY_STRING;
-        } else if (ROOT.equals(path.get(0))) {
-            return avoidPrintingPathDelimiterTwice();
+        } else if (isAbsolute()) {
```
Eirik-Eide-Saur, line 79: Its not clear to me what Absolute entails, but from det method it seems to check wheter or not it starts with "/" could the method name be changed to something like startsWithRoot?

axthosarouris: I try to follow the Path API, https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/nio/file/Path.html#isAbsolute()

svenngjeto (approved):

PR #249, opened 2021-09-15
move UnixPath from s3 package and UriWrapper from publication-api

axthosarouris:
Move UnixPath from the s3 package to core and the UriWrapper from the publication-api to commons.

The UnixPath, allows path manipulations independent of platform and the UriWrapper allows easy manipulations of URIs such as adding children or getting parent paths

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ wrapper {
 
 allprojects {
     group = 'com.github.bibsysdev'
-    version = '1.8.1'
+    version = '1.9.0'
 
     apply plugin: 'java-library'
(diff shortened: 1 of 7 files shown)


torarnet (approved):

build.gradle
```diff
@@ -25,7 +25,7 @@ wrapper {
 
 allprojects {
     group = 'com.github.bibsysdev'
-    version = '1.8.1'
+    version = '1.9.0'
```
brinxmat, line 28: Good one, can you also ping Smile about this change please

brinxmat (approved):

PR #252, opened 2021-09-23
Np2580 Using best matching media type for response

janbj:


apigateway/src/main/java/nva/commons/apigateway/RestRequestHandler.java
```diff
@@ -67,18 +70,18 @@ protected void checkHeaders(RequestInfo requestInfo) throws UnsupportedAcceptHea
...
+        return listSupportedMediaTypes().stream()
+                .filter(mediaType -> inAcceptedMediaTypeRange(mediaType, acceptMediaTypes))
+                .collect(Collectors.toList());
     }
 
-    private boolean inSupportedMediaTypeRange(MediaType mediaType) {
-        MediaType mediaTypeRange = mediaType.withoutParameters();
-        return listSupportedMediaTypes().stream()
-                .anyMatch(supportedMediaType -> supportedMediaType.is(mediaTypeRange));
+    private boolean inAcceptedMediaTypeRange(MediaType mediaType, List<MediaType> acceptMediaTypes) {
```
axthosarouris, line 79:
```
  private boolean inAcceptedMediaTypeRange(MediaType mediaType, List<MediaType> acceptMediaTypes) {
        return acceptMediaTypes.stream()
                .map(MediaType::withoutParameters)
                .anyMatch(mediaType::is);
    }
```

apigateway/src/main/java/nva/commons/apigateway/RestRequestHandler.java
```diff
@@ -89,8 +92,15 @@ private boolean inSupportedMediaTypeRange(MediaType mediaType) {
         return DEFAULT_SUPPORTED_MEDIA_TYPES;
     }
 
-    protected MediaType getDefaultResponseHeader() {
-        return listSupportedMediaTypes().get(0);
+    protected MediaType getDefaultResponseHeader(RequestInfo requestInfo) {
```
axthosarouris, line 95: getDefaultResponseHeader-> getDefaultResponseContentHeader or getDefaultResponseMediaType

axthosarouris:
```
    protected MediaType getDefaultResponseHeader(RequestInfo requestInfo) {
        if (requestInfo.getHeaders().containsKey(HttpHeaders.ACCEPT)) {
            List<MediaType> mediaTypes = parseAcceptHeader(requestInfo.getHeader(HttpHeaders.ACCEPT));
            List<MediaType> matches = findMediaTypeMatches(mediaTypes);

            return matches.stream().findFirst().orElseGet(this::defaultContentHeaderWhenNotSpecfiedByClientRequest);
        } else {
            return defaultContentHeaderWhenNotSpecfiedByClientRequest();
        }
    }

    private MediaType defaultContentHeaderWhenNotSpecfiedByClientRequest() {
        return listSupportedMediaTypes().get(0);
    }
```
orElse calculates the value always while orElseGet calculates the value only when necessary

apigateway/src/main/java/nva/commons/apigateway/RestRequestHandler.java
```diff
@@ -159,22 +169,6 @@ protected void init(OutputStream outputStream, Context context) {
...
-     * Maps an object I to an object O.
-     *
-     * @param input                 the input object of class I
-     * @param apiGatewayInputString The message of apiGateway, for extracting the headers and in case we need other
-     *                              fields during the processing
-     * @param context               the Context
-     * @return an output object of class O
-     * @throws ApiGatewayException when some predictable error happens.
-     */
-    protected O processInput(I input, String apiGatewayInputString, Context context) throws ApiGatewayException {
```
axthosarouris, line 172: This method was inlined because we need to use the requestInfo outside the method. Eitherway or another this method has never been overriden

apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandler.java
```diff
@@ -69,10 +69,10 @@ public void init(OutputStream outputStream, Context context) {
      * @throws IOException when serializing fails
      */
     @Override
-    protected void writeOutput(I input, O output)
+    protected void writeOutput(I input, O output, RequestInfo requestInfo)
         throws IOException, GatewayResponseSerializingException {
         try (BufferedWriter writer = new BufferedWriter(new OutputStreamWriter(outputStream, StandardCharsets.UTF_8))) {
-            GatewayResponse<O> gatewayResponse = new GatewayResponse<>(output, getSuccessHeaders(),
+            GatewayResponse<O> gatewayResponse = new GatewayResponse<>(output, getSuccessHeaders(requestInfo),
```
axthosarouris, line 75: export the method  calls in the GatewayResonse constructors to variables.

apigateway/src/main/java/nva/commons/apigateway/RestRequestHandler.java
```diff
@@ -89,8 +92,15 @@ private boolean inSupportedMediaTypeRange(MediaType mediaType) {
         return DEFAULT_SUPPORTED_MEDIA_TYPES;
     }
 
-    protected MediaType getDefaultResponseHeader() {
-        return listSupportedMediaTypes().get(0);
+    protected MediaType getDefaultResponseHeader(RequestInfo requestInfo) {
+        if (requestInfo.getHeaders().containsKey(HttpHeaders.ACCEPT)) {
+            List<MediaType> mediaTypes = parseAcceptHeader(requestInfo.getHeader(HttpHeaders.ACCEPT));
```
brinxmat, line 97: Probably a filtered stream

brinxmat (changes requested):

apigateway/src/main/java/nva/commons/apigateway/RestRequestHandler.java
```diff
@@ -57,7 +57,10 @@
     protected void checkHeaders(RequestInfo requestInfo) throws UnsupportedAcceptHeaderException {
```
axthosarouris, line 57: The logic of this method is nearly duplicated when we want to calculate the response that we want to send. My suggestion is  to refactor this so that the check happens when we want to calculate the result. The direct consequence of this is that one will get the UnsupportedType error only when the Response will return something, and not when failing. I do not see this as a problem, since when we are returning an error we do not necessarily return the content type the user asked.  

I have made a PR with my suggestion: https://github.com/BIBSYSDEV/nva-commons/pull/253/

axthosarouris (changes requested):

axthosarouris (approved):

brinxmat (approved):

PR #266, opened 2021-11-05
FakeS3Client lists files in order they were submitted

axthosarouris:
In order to make pagination testing easier, we force FakeS3Client to list the objects in the order they were submitted

### DIFF
--- a/nvatestutils/src/test/java/no/unit/nva/stubs/FakeS3ClientTest.java
+++ b/nvatestutils/src/test/java/no/unit/nva/stubs/FakeS3ClientTest.java
@@ -1,27 +1,38 @@
 package no.unit.nva.stubs;
 
+import static no.unit.nva.testutils.RandomDataGenerator.randomString;
 import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.collection.IsIterableContainingInOrder.contains;
 import static org.hamcrest.core.Is.is;
 import static org.hamcrest.core.IsEqual.equalTo;
 import com.github.javafaker.Faker;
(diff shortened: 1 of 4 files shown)


nvatestutils/src/test/java/no/unit/nva/stubs/FakeS3ClientTest.java
```diff
@@ -39,12 +50,51 @@ public void putObjectMakesContentAvailableForGetting() {
...
+            .map(UnixPath::of)
+            .map(UnixPath::removeRoot)
+            .map(UnixPath::toString)
+            .collect(Collectors.toList());
+    }
+
+    private ListObjectsRequest insertFilesToBucketInOrder(List<String> sampleFilenames,
+                                                          FakeS3Client s3Client) {
+        for (var filename : sampleFilenames) {
+            putObject(s3Client, URI.create(SOME_BUCKET_URI + "/" + filename), randomString());
```
axthosarouris, line 82: I do not want to use any tools that will create circular dependencies.

s3/src/main/java/no/unit/nva/s3/S3Driver.java
```diff
@@ -135,6 +135,12 @@ public URI insertAndCompressFiles(UnixPath s3Folder, List<String> content) throw
         return insertAndCompressObjects(s3Folder, content);
     }
 
+    @JacocoGenerated
```
axthosarouris, line 138: moved to be next to its overloaded method

brinxmat (approved):

PR #272, opened 2021-11-16
EventBody does not require topic setter

axthosarouris:


eventhandlers/src/main/java/no/unit/nva/events/models/EventBody.java
```diff
@@ -8,6 +8,4 @@
 
     @JsonProperty(TOPIC)
     String getTopic();
-
```
axthosarouris, line 11: Interface for EventBodies that contain a topic. More often than not we use constructors as json creators and setters are not necessary.

brinxmat (approved):

PR #273, opened 2021-11-16
upgrade to gradle 7.3 (wip)

axthosarouris:
This is the first PR in order to migrate nva-commons to the new style. 
Actions:
1. Upgrade to gradle 7,
2. Add the plugin files (copied from nva-filemodel) but do not use
3. Update the publish plugin module with the correct information.


buildSrc/src/main/groovy/nvacommons.jacoco-merge.gradle
```diff
@@ -0,0 +1,62 @@
+//For generating one unified coverage report
+def getProjectList() {
```
axthosarouris, line 2: These files are not in use right now. The will be used in subsequent PR.

PR #288, opened 2022-01-05
alpha: refactoring convention file attempting to reduce dependencies

axthosarouris:


apigateway/build.gradle
```diff
@@ -4,15 +4,27 @@ plugins {
 }
 
 dependencies {
+    api(libs.aws.lambda.core)
+
     implementation project(":core")
     implementation libs.guava
+    implementation libs.bundles.logging
+
+    implementation libs.jackson.core
```
axthosarouris, line 13: avoid using the jackson bundle in order to be able to remove easier unused dependecies

eventhandlers/build.gradle
```diff
@@ -4,10 +4,15 @@ plugins {
 }
 
 dependencies {
+    api(libs.aws.lambda.core)
```
axthosarouris, line 7: A very small (10 classes) but essential library for all lambdas. Easier to have it included by default in the client code

buildSrc/src/main/groovy/nvacommons.java-conventions.gradle
```diff
@@ -9,7 +9,8 @@ plugins {
 }
 
 group 'com.github.bibsysdev'
-version = '1.16.30'
+version = '1.16.31-alpha'
```
axthosarouris, line 12: alpha because it would be better to test it first in dev and sandbox and verify that everything is working as it should

janbj (approved):

brinxmat (approved):

PR #299, opened 2022-01-18
NP-3544: move ScanDatabaseRequest from publication service to commons for reuse

axthosarouris:
ScanDatabaseRequest is a class that is used currenty in Publication service for scanning/migrating/refreshing the publication. We need the same functionality in identity service and I thought we could reuse this event

### DIFF
--- a/eventhandlers/build.gradle
+++ b/eventhandlers/build.gradle
@@ -10,6 +10,9 @@ dependencies {
     implementation libs.aws.sdk2.core
     implementation libs.aws.sdk.core
     implementation libs.jackson.databind
+    implementation libs.jackson.core
+    implementation libs.aws.sdk.dynamo
+    implementation libs.aws.sdk2.eventbridge
 
     testImplementation libs.bundles.testing
     testImplementation libs.bundles.logging
--- a/eventhandlers/src/main/java/no/unit/nva/events/models/ScanDatabaseRequest.java
+++ b/eventhandlers/src/main/java/no/unit/nva/events/models/ScanDatabaseRequest.java
@@ -12,6 +12,12 @@ import nva.commons.core.JacocoGenerated;
 import nva.commons.core.JsonSerializable;
 import software.amazon.awssdk.services.eventbridge.model.PutEventsRequestEntry;
 
+/**
+ * Class that can be sent as an event to a lambda handler for scanning a page of a DynamoDB table. The {@code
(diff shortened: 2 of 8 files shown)


eventhandlers/build.gradle
```diff
@@ -10,6 +10,9 @@ dependencies {
     implementation libs.aws.sdk2.core
     implementation libs.aws.sdk.core
     implementation libs.jackson.databind
+    implementation libs.jackson.core
```
axthosarouris, line 13: necessary libs for the new class

eventhandlers/src/main/java/no/unit/nva/events/models/ScanDatabaseRequest.java
```diff
@@ -0,0 +1,104 @@
...
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import java.time.Instant;
+import java.util.Map;
+import java.util.Objects;
+import nva.commons.core.JacocoGenerated;
+import nva.commons.core.JsonSerializable;
+import software.amazon.awssdk.services.eventbridge.model.PutEventsRequestEntry;
+
+public class ScanDatabaseRequest implements JsonSerializable {
```
axthosarouris, line 15: copied from publication service

gradle/libs.versions.toml
```diff
@@ -39,8 +39,8 @@ aws-sdk2-secrets = { group = 'software.amazon.awssdk', name = 'secretsmanager',
 aws-sdk2-urlconnection = { group = 'software.amazon.awssdk', name = 'url-connection-client', version.ref = 'awsSdk2' }
 
 aws-ion = { group = 'software.amazon.ion', name = 'ion-java', version.ref = 'awsIon' }
-aws-sdk-s3 = { group = 'software.amazon.awssdk', name = 's3', version.ref = 'awsSdk2' }
-aws-sdk-eventbridge = { group = 'software.amazon.awssdk', name = 'eventbridge', version.ref = 'awsSdk2' }
+aws-sdk2-s3 = { group = 'software.amazon.awssdk', name = 's3', version.ref = 'awsSdk2' }
```
axthosarouris, line 42: renamed to sdk2 to distinguish from the sdk-1 packages. The target is that the sdk-1 packages should disappear

brinxmat (approved):

PR #303, opened 2022-01-24
NP-3566: https is the default scheme

axthosarouris:


core/src/main/java/nva/commons/core/paths/UriWrapper.java
```diff
@@ -111,4 +125,9 @@ public UriWrapper addQueryParameters(Map<String, String> parameters) {
     public String toString() {
         return this.getUri().toString();
     }
+
+    private static URI createUriWithSchemeAndHost(String scheme, String host) {
```
axthosarouris, line 129: moved from line 35

PR #312, opened 2022-02-01
feature: scan-database-request for sdk2

axthosarouris:
This is basically a cloning of the class ScanDatabaseRequest to create the ScanDatabaseRequestV2 for the sdk2. It is necessary for the migration of the identity service from sdk1 to sdk2.

### DIFF
--- a/eventhandlers/src/main/java/no/unit/nva/events/models/AbstractScanDatabaseRequest.java
+++ b/eventhandlers/src/main/java/no/unit/nva/events/models/AbstractScanDatabaseRequest.java
@@ -1,0 +1,113 @@
+package no.unit.nva.events.models;
+
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.time.Instant;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import no.unit.nva.commons.json.JsonSerializable;
+import nva.commons.core.JacocoGenerated;
+import software.amazon.awssdk.services.eventbridge.model.PutEventsRequestEntry;
+
+/**
+ * Class that can be sent as an event to a lambda handler for scanning a page of a DynamoDB table. The {@code
+ * startMarker} is a scan start marker as required by the DynamoDb client. The {@code pageSize} is the number of the
+ * results the scan will return (max 1000). The {@code topic} is the event topic that the handler is listening for
+ * events.
+ */
+public abstract class AbstractScanDatabaseRequest<T> implements EventBody, JsonSerializable {
+
+    public static final String START_MARKER = "startMarker";
+    public static final String PAGE_SIZE = "pageSize";
+
+    public static final int DEFAULT_PAGE_SIZE = 700; // Choosing for safety 3/4 of max page size.
+    public static final int MAX_PAGE_SIZE = 1000;
+
+    @JsonProperty(START_MARKER)
+    private final Map<String, T> startMarker;
+    @JsonProperty(PAGE_SIZE)
+    private final int pageSize;
+    @JsonProperty(TOPIC)
+    private final String topic;
+
+    @JsonCreator
+    public AbstractScanDatabaseRequest(
+        @JsonProperty(TOPIC) String topic,
+        @JsonProperty(PAGE_SIZE) Integer pageSize,
+        @JsonProperty(START_MARKER) Map<String, T> startMarker) {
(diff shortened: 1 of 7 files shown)


eventhandlers/src/main/java/no/unit/nva/events/models/AbstractScanDatabaseRequest.java
```diff
@@ -0,0 +1,113 @@
...
+import nva.commons.core.JacocoGenerated;
+import software.amazon.awssdk.services.eventbridge.model.PutEventsRequestEntry;
+
+/**
+ * Class that can be sent as an event to a lambda handler for scanning a page of a DynamoDB table. The {@code
+ * startMarker} is a scan start marker as required by the DynamoDb client. The {@code pageSize} is the number of the
+ * results the scan will return (max 1000). The {@code topic} is the event topic that the handler is listening for
+ * events.
+ */
+public abstract class AbstractScanDatabaseRequest<T> implements EventBody, JsonSerializable {
```
axthosarouris, line 20: This is the old ScanDatabaseRequest with the only difference that is now a generic class inherited by the ScanDatabaseRequest and the ScanDatabaseRequestV2

eventhandlers/src/main/java/no/unit/nva/events/models/ScanDatabaseRequest.java
```diff
@@ -5,114 +5,25 @@
...
-        this.startMarker = startMarker;
-        this.topic = topic;
+    public ScanDatabaseRequest(@JsonProperty(TOPIC) String topic,
+                               @JsonProperty(PAGE_SIZE) Integer pageSize,
+                               @JsonProperty(START_MARKER) Map<String, AttributeValue> startMarker) {
+        super(topic, pageSize, startMarker);
     }
 
-    public static ScanDatabaseRequest fromJson(String detail) throws JsonProcessingException {
+    public static ScanDatabaseRequest fromJson(String detail)
```
axthosarouris, line 25: The static method is declared here and not in the abstract class, for compatibility reasons. Also, if it was created in the abstract class the generated instances would be AbstractScanDatabaseRequest<AttributeValue> which is slightly different from ScanDatabaseRequest.( A ScanDatabaseRequest is an AbstractScanDatabaseRequest<AttributeValue>, but the opposite is not true)

eventhandlers/src/main/java/no/unit/nva/events/models/ScanDatabaseRequestV2.java
```diff
@@ -0,0 +1,29 @@
...
+import java.util.Map;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+
+/**
+ * Class that can be sent as an event to a lambda handler for scanning a page of a DynamoDB table. The {@code
+ * startMarker} is a scan start marker as required by the DynamoDb client. The {@code pageSize} is the number of the
+ * results the scan will return (max 1000). The {@code topic} is the event topic that the handler is listening for
+ * events.
+ */
+public class ScanDatabaseRequestV2 extends AbstractScanDatabaseRequest<AttributeValue> {
```
axthosarouris, line 16: The difference is that this AttributeValue is the sdk2 version of the AttributeValue.

eventhandlers/src/test/java/no/unit/nva/events/models/ScanDatabaseRequestV2Test.java
```diff
@@ -0,0 +1,63 @@
...
+import static org.hamcrest.core.IsEqual.equalTo;
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.util.Map;
+import no.unit.nva.commons.json.JsonUtils;
+import no.unit.nva.events.EventsConfig;
+import org.junit.jupiter.api.Test;
+
+class ScanDatabaseRequestV2Test {
```
axthosarouris, line 16: Basically a copy past of the tests of the ScanDatabaseRequest, because I could not find a way to parametrize the tests.

eventhandlers/src/test/java/no/unit/nva/events/models/ScanDatabaseRequestV2Test.java
```diff
@@ -0,0 +1,63 @@
...
+
+    public static final String EVENT_TOPIC = "topic";
+
+    @Test
+    void shouldReturnAnEventBridgeEventWhereTheTopicIsSetInTheDetailBody() throws JsonProcessingException {
+        String expectedTopic = randomString();
```
brinxmat, line 22: String

brinxmat (approved):

PR #315, opened 2022-02-02
NP-3574: event can have subtopic

axthosarouris:


eventhandlers/src/main/java/no/unit/nva/events/models/EventReference.java
```diff
@@ -18,22 +18,36 @@
...
     @JsonCreator
     public EventReference(@JsonProperty(TOPIC) String topic,
+                          @JsonProperty(SUBTOPIC) String subtopic,
                           @JsonProperty(URI) URI uri) {
         this.topic = topic;
+        this.subtopic = subtopic;
         this.uri = uri;
     }
 
+    public EventReference(String topic, URI uri) {
```
axthosarouris, line 38: constructor for keeping compatibility the the current code. also subtopic is not always necessary

brinxmat (approved):

PR #316, opened 2022-02-02
NP-3574: SDK2 AttributeValue is not serializable

axthosarouris:
SDK2 AttributeValue is not serializable and ScanDatabaseRequestV2 is not serialized properly. I have implemented a workaround where the start marker is stored as a String,String Map and is converted on demand by specialized methods.


eventhandlers/src/main/java/no/unit/nva/events/models/ScanDatabaseRequest.java
```diff
@@ -5,30 +5,115 @@
...
+import software.amazon.awssdk.services.eventbridge.model.PutEventsRequestEntry;
 
 /**
  * Class that can be sent as an event to a lambda handler for scanning a page of a DynamoDB table. The {@code
  * startMarker} is a scan start marker as required by the DynamoDb client. The {@code pageSize} is the number of the
  * results the scan will return (max 1000). The {@code topic} is the event topic that the handler is listening for
  * events.
  */
-public class ScanDatabaseRequest extends AbstractScanDatabaseRequest<AttributeValue> {
+public class ScanDatabaseRequest implements EventBody, JsonSerializable {
```
axthosarouris, line 22: This class was restored to be as it was before. Inheritance complicates things and this class will not survive for long

eventhandlers/src/main/java/no/unit/nva/events/models/ScanDatabaseRequestV2.java
```diff
@@ -1,34 +1,126 @@
...
+import software.amazon.awssdk.services.eventbridge.model.PutEventsRequestEntry;
 
 /**
  * Class that can be sent as an event to a lambda handler for scanning a page of a DynamoDB table. The {@code
  * startMarker} is a scan start marker as required by the DynamoDb client. The {@code pageSize} is the number of the
  * results the scan will return (max 1000). The {@code topic} is the event topic that the handler is listening for
  * events.
  */
-public class ScanDatabaseRequestV2 extends AbstractScanDatabaseRequest<AttributeValue> {
+public class ScanDatabaseRequestV2 implements EventBody, JsonSerializable {
```
axthosarouris, line 25: Basically a copy paste of the ScanDatabaseRequest, with the only difference that the marker is a String,String map. This s because SDK2 AttributeValue is not serializable.

brinxmat (approved):

PR #320, opened 2022-02-13
Apigatewayv2

axthosarouris:
A simpler and hopefully slightly faster ApiGatewayHandler that satisfies (almost) all requirements of ApiGatewayHandler as they are expressed by the tests. The only requirements that are not satisfied are the ones that have to do with deserialization of the request body. This is because the responsibility of this will be moved to the implementing methods. 

This is not an effort to substitute Micronaut. This is an effort to check whether we can make some handlers faster until we migrate to Micronaut without changing the code of the services too much.

The major changes are the following:
1. We do not parse the ApiGatewayRequestEvents. Amazon does this for us.
2. This ApiGatewayHandlerV2 is intended to be independent of serialization libraries or in worst case to use something very light. This will come in next PR after we have established the the handler works.

### DIFF
--- a/apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandlerV2.java
+++ b/apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandlerV2.java
@@ -1,0 +1,112 @@
+package nva.commons.apigateway;
+
+import static com.google.common.net.HttpHeaders.ACCESS_CONTROL_ALLOW_ORIGIN;
+import static com.google.common.net.MediaType.JSON_UTF_8;
+import static java.net.HttpURLConnection.HTTP_INTERNAL_ERROR;
+import static java.util.Objects.isNull;
+import static no.unit.nva.commons.json.JsonUtils.dtoObjectMapper;
+import static nva.commons.core.attempt.Try.attempt;
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.TreeTraversingParser;
+import com.google.common.net.MediaType;
+import java.io.IOException;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Supplier;
+import no.unit.nva.commons.json.JsonUtils;
+import nva.commons.apigateway.exceptions.ApiGatewayException;
+import nva.commons.core.Environment;
+import org.zalando.problem.Problem;
+import org.zalando.problem.Status;
+import org.zalando.problem.ThrowableProblem;
+
+public abstract class ApiGatewayHandlerV2<I, O>
+    implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
+
+    private static final List<MediaType> DEFAULT_SUPPORTED_MEDIA_TYPES = List.of(JSON_UTF_8);
+    private static final String ALLOWED_ORIGIN = new Environment().readEnv("ALLOWED_ORIGIN");
+    private Class<I> iclass;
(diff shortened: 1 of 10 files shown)


apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandlerV2.java
```diff
@@ -0,0 +1,169 @@
...
+import nva.commons.core.Environment;
+import nva.commons.core.attempt.Try;
+import nva.commons.core.exceptions.ExceptionUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.zalando.problem.Problem;
+import org.zalando.problem.Status;
+import org.zalando.problem.ThrowableProblem;
+
+public abstract class ApiGatewayHandlerV2<I, O>
```
axthosarouris, line 37: Some of the functionality of the ApiGatewayHandler has never been used. I have kept it however, because I did not want to omit any tests. We can refine the handler later.

apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandlerV2.java
```diff
@@ -0,0 +1,169 @@
...
+
+    protected I parseInput(String body) throws JsonProcessingException {
+        if (isNull(body)) {
+            return null;
+        } else if (String.class.equals(iclass)) {
+            return (I) body;
+        } else {
+            return dtoObjectMapper.readValue(body, iclass);
+        }
+    }
```
axthosarouris, line 99: This will be removed in next iteration

apigateway/src/main/java/nva/commons/apigateway/MediaTypes.java
```diff
@@ -1,14 +1,71 @@
...
     public static final MediaType APPLICATION_PROBLEM_JSON = MediaType.create("application", "problem+json");
-
+    public static final String ACCEPT_HEADER_VALUES_DELIMITER = ",";
+    public static final List<MediaType> DEFAULT_SUPPORTED_MEDIA_TYPES = List.of(JSON_UTF_8.withoutParameters());
+    public static final int MOST_PREFERRED_DEFAULT_MEDIA_TYPE = 0;
 
     private MediaTypes() {
     }
 
+    public static MediaType parse(String mediaType, List<MediaType> supportedMediaTypes)
```
axthosarouris, line 25: moved the logic of parsing the Accept headers here. All tests pass.

apigateway/src/test/java/nva/commons/apigateway/ApiGatewayHandlerV2Test.java
```diff
@@ -0,0 +1,442 @@
...
+import nva.commons.logutils.TestAppender;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+import org.zalando.problem.Problem;
+import org.zalando.problem.Status;
+
+public class ApiGatewayHandlerV2Test {
```
axthosarouris, line 54: All tests are the same as the ones in ApiGatewayHandlerTest

apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandlerV2.java
```diff
@@ -0,0 +1,169 @@
...
+
+    private BadRequestException handleInputParsingError(Try<I> fail) {
+        return new BadRequestException(fail.getException().getMessage(), fail.getException());
+    }
+
+    private ThrowableProblem createProblem(String message, Integer statusCode, Context context) {
+        Status status = Status.valueOf(statusCode);
```
brinxmat, line 106: Any reason why this isn't inline?

apigateway/src/main/java/nva/commons/apigateway/MediaTypes.java
```diff
@@ -1,14 +1,71 @@
...
+        throws UnsupportedAcceptHeaderException {
+        var mediaTypeString = StringUtils.removeWhiteSpaces(mediaType);
+        MediaType mostPreferredMediaType = mostPreferredMediaType(supportedMediaTypes);
```
brinxmat, line 28: var?

brinxmat (dismissed):

apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandlerV2.java
axthosarouris: I use it twice

brinxmat (approved):

PR #347, opened 2022-03-25
RequestInfo throws Unauthoriized by default when user cannot be idetified

axthosarouris:
…ified


apigateway/src/main/java/nva/commons/apigateway/RequestInfo.java
```diff
@@ -239,14 +240,17 @@ public boolean userIsAuthorized(String accessRight) {
     }
 
     @JsonIgnore
-    public Optional<String> getCustomerId() {
+    public String getCustomerId() throws UnauthorizedException {
```
axthosarouris, line 243: Initially we said it should be the application that decides, but more often than not if a client needs it, then it cannot do much without it.

PR #352, opened 2022-04-13
Authorized backend client does not read env variables for backend access token when user access token is used

axthosarouris:


auth/src/test/java/no/unit/nva/auth/AuthorizedBackendClientTest.java
```diff
@@ -72,4 +76,11 @@ void shouldSendRequestsContainingTheUserAccessTokenWhenUserAccessTokenIsSubmitte
             client.sendAsync(request, BodyHandlers.ofString(StandardCharsets.UTF_8)).join();
         assertThat(asyncResponse.body(), containsString(protectedContent));
     }
+
+    @Test
+    void shouldNotRequireEnvVariablesForBackendAccessTokenIfUserAccessTokenIsUsed() {
+        assertThat(CLIENT_ID, is(instanceOf(Supplier.class)));
```
axthosarouris, line 82: Ideally the test would be a test where we unset the variables and we verify that the client works, but this means that we make the environment a non static field and inject it through the constructor. Similarly all  static variables should be non-static. I was tempted to do this, but it is quite a refactoring for very little gain

auth/src/test/java/no/unit/nva/auth/AuthorizedBackendClientTest.java
```diff
@@ -72,4 +79,22 @@ void shouldSendRequestsContainingTheUserAccessTokenWhenUserAccessTokenIsSubmitte
...
+        assertThat(CLIENT_SECRET, is(instanceOf(Supplier.class)));
+        assertThat(COGNITO_URI, is(instanceOf(Supplier.class)));
+    }
+
+    @Test
+    void shouldReadAllNecessaryInformationFromEnvWhenBackendAccessTokenIsUsed() {
+        Executable action = AuthorizedBackendClient::prepareWithBackendCredentials;
```
axthosarouris, line 92: This is mainly because the Jacoco complains about the suppliers.

brinxmat (approved):

PR #360, opened 2022-04-28
AuthorizedBackendClient accepts CognitoCredentials object

axthosarouris:
Remove the requirement for reading Cognito Credentials from env variables.

Reading from Env variables makes this more difficult because and has implications with deployment when the secrets are updated.


gregorgabriel (approved):
Good job, thanks.

PR #364, opened 2022-05-03
refresh the access token before every request

axthosarouris:


auth/src/main/java/no/unit/nva/auth/AuthorizedBackendClient.java
```diff
@@ -60,18 +61,24 @@ public static AuthorizedBackendClient prepareWithBearerToken(HttpClient httpClie
...
 
     @JacocoGenerated
-    public String getBearerToken() {
+    protected String getBearerToken() {
         return bearerToken;
     }
 
     public <T> HttpResponse<T> send(HttpRequest.Builder request, BodyHandler<T> responseBodyHandler)
         throws IOException, InterruptedException {
+        if (bearerTokenIsNotInjectedDirectly) {
```
brinxmat, line 70: Do you ever refresh the token without this check?

brinxmat (approved):
You might consider the comment in another PR

auth/src/main/java/no/unit/nva/auth/AuthorizedBackendClient.java
axthosarouris: No, but if the token is set during initialization, i.e., the token is not acquired through Cognito (e.g. during testing) then we cannot/should not try to refresh it.

PR #366, opened 2022-05-05
allow Unauthorized to have a message and HandleRequestBuilder creates requests for ApiGatewayv2

axthosarouris:


nvatestutils/src/main/java/no/unit/nva/testutils/HandlerRequestBuilder.java
```diff
@@ -127,6 +127,11 @@ public InputStream build() throws JsonProcessingException {
         return new ByteArrayInputStream(objectMapper.writeValueAsBytes(this));
     }
 
+    public APIGatewayProxyRequestEvent buildRequestEvent() throws JsonProcessingException {
```
axthosarouris, line 130: Create a method for building requests for apigatewayv2.

brinxmat (approved):

PR #368, opened 2022-05-05
Request info check for app admin

axthosarouris:


torarnet (approved):
Do we need a Access Right for Editor or can we use Application Admin?

PR #371, opened 2022-05-06
Replace COGNITO_URI with COGNITO_HOST

axthosarouris:


apigateway/src/main/java/nva/commons/apigateway/RequestInfo.java
```diff
@@ -2,11 +2,28 @@
 
 import static java.util.Objects.isNull;
 import static java.util.function.Predicate.not;
-import static no.unit.nva.auth.CognitoUserInfo.NVA_USERNAME_CLAIM;
-import static no.unit.nva.auth.CognitoUserInfo.PERSON_CRISTIN_ID_CLAIM;
-import static no.unit.nva.auth.CognitoUserInfo.TOP_LEVEL_ORG_CRISTIN_ID_CLAIM;
+import static nva.commons.apigateway.RequestInfoConstants.AUTHORIZATION_FAILURE_WARNING;
```
axthosarouris, line 5: moved all constants in a different class

PR #374, opened 2022-05-12
Np 9117 add nin to requestinfo

torarnet:


apigateway/src/main/java/nva/commons/apigateway/RequestInfo.java
```diff
@@ -269,27 +271,36 @@ public Optional<URI> getTopLevelOrgCristinId() {
...
     public URI getPersonCristinId() throws UnauthorizedException {
         return extractPersonCristinIdOffline()
-            .or(this::fetchPersonCristinIdFromCognito)
-            .orElseThrow(UnauthorizedException::new);
+                   .or(this::fetchPersonCristinIdFromCognito)
+                   .orElseThrow(UnauthorizedException::new);
+    }
+
+    @JsonIgnore
+    public String getPersonNinId() {
```
axthosarouris, line 287: getPersonNin, not NinId

auth/src/main/java/no/unit/nva/auth/CognitoUserInfo.java
```diff
@@ -30,6 +32,10 @@ public class CognitoUserInfo {
     private URI topOrgCristinId;
     @JsonProperty(PERSON_CRISTIN_ID_CLAIM)
     private URI personCristinId;
+    @JsonProperty(PERSON_NIN_ID_CLAIM)
```
axthosarouris, line 35: Use the @JsonAlias to avoid having two fields

nvatestutils/src/main/java/no/unit/nva/testutils/HandlerRequestBuilder.java
```diff
@@ -200,6 +202,18 @@ public HandlerRequestBuilder<T> withPersonCristinId(URI personCristinId) {
         return this;
     }
 
+    public HandlerRequestBuilder<T> withPersonNinId(String personNinId) {
```
axthosarouris, line 205: Here we start to leak cognito specific things to everywhere. We should have one method here, `withPersonNin`

apigateway/src/main/java/nva/commons/apigateway/RequestInfo.java
```diff
@@ -321,15 +332,35 @@ private Optional<URI> extractPersonCristinIdOffline() {
...
+                   .map(CognitoUserInfo::getPersonNinId)
+                   .toOptional();
+    }
+
+    private Optional<String> extractPersonFeideNinIdOffline() {
+        return getRequestContextParameterOpt(PERSON_FEIDE_NIN_ID);
+    }
+
+    private Optional<String> extractPersonNinIdOffline() {
+        return getRequestContextParameterOpt(PERSON_NIN_ID);
```
axthosarouris, line 356: Since here we are only trying to inject a value, we do not have to have two ways to inject a value. Use the "custom:nin" claim only everywhere, except for CognitoInfo where you should add the alias as directed

axthosarouris (changes requested):
Use only one claim `custom:nin` and the other one use it only as a JsonAlias when you are reading the Cognito userinfo data.

auth/src/main/java/no/unit/nva/auth/CognitoUserInfo.java
```diff
@@ -17,6 +18,8 @@ public class CognitoUserInfo {
     public static final String NVA_USERNAME_CLAIM = "custom:nvaUsername";
     public static final String TOP_LEVEL_ORG_CRISTIN_ID_CLAIM = "custom:topOrgCristinId";
     public static final String PERSON_CRISTIN_ID_CLAIM = "custom:cristinId";
+    public static final String PERSON_NIN_CLAIM = "custom:nin";
```
svenngjeto, line 21: NIN is the same wherever i comes from, shouldn't  "custom:nin" and  "custom:feideIdNin" be just one claim?

axthosarouris: These information come from dataporten. They are different fields in dataporten and in Cognito we cannot map two input fields (Dataporten) to the same cognito field.

auth/src/main/java/no/unit/nva/auth/CognitoUserInfo.java
```diff
@@ -30,6 +33,9 @@ public class CognitoUserInfo {
     private URI topOrgCristinId;
     @JsonProperty(PERSON_CRISTIN_ID_CLAIM)
     private URI personCristinId;
+    @JsonAlias({PERSON_NIN_CLAIM, PERSON_FEIDE_NIN_CLAIM})
```
axthosarouris, line 36: You need to include only the "PERSON_FEIDE_NIN_CLAIM" in the alias.

axthosarouris (approved):

brinxmat (approved):

axthosarouris (approved):

PR #381, opened 2022-06-03
read user info from alternative endpoint

axthosarouris:


apigateway/src/test/java/nva/commons/apigateway/RequestInfoTest.java
```diff
@@ -582,41 +611,6 @@ private RequestInfo extractAccessRightsFromApiGatewayEvent()
         return apiMessageParser.getRequestInfo(event);
     }
 
-    private List<SimpleEntry<String, ObjectNode>> createNodesForEachPathElement(JsonPointer jsonPointer) {
```
axthosarouris, line 585: this was dead code

brinxmat (approved):

PR #389, opened 2022-08-28
Preliminary exploration of what documentation is required

brinxmat:
Here, I'm attempting to add documentation for the top-level and add a lower-level "main" doc, I think it is useful at the top level to link to the main docs for each module. I think it is at the top level that we should say generally what things do and which. to choose if there is a choice (e.g. apigw v1 or v2). I wonder if adding the install information here is also useful for quick reference? Maybe overkill.

In the main docs, I try to provide examples, but some of the docs will become huge and a bit unhelpful since, e.g., RequestInfo is a POJO with many methods, and here I think it is more helpful to document what the methods do and what they return…not sure I did an amazing job here, but it is a start.

I recognize that I struggle to explain a few things and have added notes for the reviewer, these should be removed in the final version by a valid explanation. Maybe I should be more descriptive in general too?

I also see we have a number of visibility issues, I have remarked on these and we should have a discussion before we move forward.

### DIFF
--- a/apigateway/README.md
+++ b/apigateway/README.md
@@ -1,0 +1,194 @@
+# ApiGateway
+
+This library provides Java utilities that provide a uniform way of working with AWS API gateway events in AWS Lambdas.
+
+#### Install
+
+```groovy
+runtimeOnly group: 'com.github.bibsysdev', name: 'apigateway', version: '$version'
+```
+
+### AccessRight
+
+This enum provides the known access rights in the NVA platform.
+
+### AccessRightEntry
+
+Should this be an exposed API at all? It is used by RequestInfo, HandlerRequestBuilder, but is never accessed directly.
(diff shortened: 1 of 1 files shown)


apigateway/README.md
```diff
@@ -0,0 +1,194 @@
+# ApiGateway
+
+This library provides Java utilities that provide a uniform way of working with AWS API gateway events in AWS Lambdas.
+
+#### Install
+
+```groovy
+runtimeOnly group: 'com.github.bibsysdev', name: 'apigateway', version: '$version'
```
axthosarouris, line 8: usually we need more that runtimeOnly because we extend the ApiGatewayHandler and we use the ApiGateway Exceptions

axthosarouris (approved):

PR #392, opened 2022-09-12
NP-13048 added support for reading plain-text secret as JSON

ketilaa:


brinxmat (dismissed):

nvatestutils/src/main/java/no/unit/nva/stubs/FakeSecretsManagerClient.java
```diff
@@ -27,15 +28,31 @@ public FakeSecretsManagerClient putSecret(String name, String key, String value)
...
         return Optional.ofNullable(getSecretValueRequest.secretId())
                    .map(SecretName::new)
-                   .map(secretName -> secrets.get(secretName))
+                   .flatMap(this::resolveSecret)
                    .map(secretContents -> addSecretContents(secretContents, getSecretValueRequest))
                    .orElseThrow();
     }
 
+    private Optional<String> resolveSecret(SecretName secretName) {
+        if (secrets.containsKey(secretName)) {
```
ketilaa, line 47: Tar gjerne innspill på om det går an å gjøre dette mer elegant.

nvatestutils/src/test/java/no/unit/nva/stubs/FakeSecretsManagerClientTest.java
```diff
@@ -24,4 +28,31 @@ void shouldBeUsableWithSecretsReader() {
         assertThat(secretsReader.fetchSecret(secretName, secretKey1), is(equalTo(secretValue1)));
         assertThat(secretsReader.fetchSecret(secretName, secretKey2), is(equalTo(secretValue2)));
     }
+
+    @Test
+    void shouldBeUsableWithSecretsReaderUsingPlainTextJson() {
```
axthosarouris, line 33: since the secret in AWS is considered to be a plain text, then maybe it is better for the SecretsReader to return a String and the responibility to parse to JSON to be in the application (i.e. the code that is interested in the secret value).

example:
```
var secretString = secretsReader.fetchPlainTextSecret(secretName);
var json = parseJson(secretString)
```

secrets/src/test/java/nva/commons/secrets/SecretsReaderTest.java
```diff
@@ -68,6 +72,12 @@ public void fetchSecretReturnsSecretValueWhenSecretNameAndSecretKeyAreCorrect()
         assertThat(value, is(equalTo(SECRET_VALUE)));
     }
 
+    @Test
+    public void fetchJsonSecretReturnsDeserializedObjectCorrectly() {
```
axthosarouris, line 76: it should not be the responsibilty of the secrets reader to parse the value of a plain text secret. What if the application chooses to use another type of serialization? we shouldn't need to change commons for that,

axthosarouris (changes requested):
Good job, Not sure about the necessity JSON parsing

nvatestutils/src/test/java/no/unit/nva/stubs/FakeSecretsManagerClientTest.java
ketilaa: Fixed

secrets/src/test/java/nva/commons/secrets/SecretsReaderTest.java
ketilaa: Fixed

axthosarouris (approved):

PR #393, opened 2022-09-15
Added ApiGatewayProxyHandler which extends ApiGatewayHandler but isnt…

sondrev:
…ead of abstracting two different methods; one for body and one for http-status, it only abstracts one which needs to return both.

### DIFF
--- a/apigateway/src/main/java/nva/commons/apigateway/ApiGatewayProxyHandler.java
+++ b/apigateway/src/main/java/nva/commons/apigateway/ApiGatewayProxyHandler.java
@@ -1,0 +1,66 @@
+package nva.commons.apigateway;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import nva.commons.apigateway.exceptions.ApiGatewayException;
+import nva.commons.core.Environment;
+import nva.commons.core.JacocoGenerated;
+import org.apache.commons.lang3.tuple.Pair;
+
+/**
+ * An extension of ApiGatewayHandler where you are also able to specify the http status-code.
+ *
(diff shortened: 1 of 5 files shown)


apigateway/src/main/java/nva/commons/apigateway/ApiGatewayProxyHandler.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @JacocoGenerated
+    public ApiGatewayProxyHandler(Class<I> iclass, Environment environment, ObjectMapper objectMapper) {
+        super(iclass, environment, objectMapper);
+    }
+
+    @Override
+    protected O processInput(I input, RequestInfo requestInfo, Context context) throws ApiGatewayException {
+        Pair<O, Integer> result = processProxyInput(input, requestInfo, context);
```
axthosarouris, line 37: Lambda start up times are affected by the number of classes we are loading through dependencies. We should avoid depending on a whole module for a simple class. 
Using Pair is maybe a bit of an anti-pattern here because we always know what left and right is here and we should define them in a (ProxyResponse) class. That way we make our code more robust against wrong usage (e.g. someone mixing left with right). 
See https://github.com/BIBSYSDEV/nva-commons/pull/394 for a suggestion.

axthosarouris (changes requested):

apigateway/src/main/java/nva/commons/apigateway/ApiGatewayProxyHandler.java
sondrev: Agreed, and thank you for providing a fix which I merged!

axthosarouris (approved):

PR #405, opened 2022-11-02
Fix media type handling for schema org

brinxmat:
- Adds support for the mediatype used for schema.org representations by DataCite
- Improves the testing for mediatypes
- Fixes errors occurring during build due to lacking jackson extensions

### DIFF
--- a/apigateway/src/main/java/nva/commons/apigateway/MediaTypes.java
+++ b/apigateway/src/main/java/nva/commons/apigateway/MediaTypes.java
@@ -4,13 +4,14 @@ import com.google.common.net.MediaType;
 
 public final class MediaTypes {
 
-    public static final MediaType APPLICATION_JSON_LD = MediaType.create("application", "ld+json");
+    public static final String APPLICATION = "application";
+    public static final MediaType APPLICATION_JSON_LD = MediaType.create(APPLICATION, "ld+json");
     public static final MediaType APPLICATION_PROBLEM_JSON =
-        MediaType.create("application", "problem+json");
+        MediaType.create(APPLICATION, "problem+json");
     public static final MediaType APPLICATION_DATACITE_XML =
-        MediaType.create("application", "vnd.datacite.datacite+xml");
+        MediaType.create(APPLICATION, "vnd.datacite.datacite+xml");
     public static final MediaType SCHEMA_ORG =
-        MediaType.create("application", "vnd.schemaorg.ld+json");
+        MediaType.create(APPLICATION, "vnd.schemaorg.ld+json");
 
     private MediaTypes() {
     }
--- a/apigateway/src/main/java/nva/commons/apigateway/MediaTypes.java
+++ b/apigateway/src/main/java/nva/commons/apigateway/MediaTypes.java
@@ -9,6 +9,8 @@ public final class MediaTypes {
         MediaType.create("application", "problem+json");
     public static final MediaType APPLICATION_DATACITE_XML =
         MediaType.create("application", "vnd.datacite.datacite+xml");
+    public static final MediaType SCHEMA_ORG =
+        MediaType.create("application", "vnd.schemaorg.ld+json");
 
     private MediaTypes() {
     }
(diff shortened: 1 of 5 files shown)


apigateway/src/main/java/nva/commons/apigateway/MediaTypes.java
```diff
@@ -4,11 +4,14 @@
 
 public final class MediaTypes {
 
-    public static final MediaType APPLICATION_JSON_LD = MediaType.create("application", "ld+json");
+    public static final String APPLICATION = "application";
```
brinxmat, line 7: PMD complains because the word is repeated four times

apigateway/src/test/java/nva/commons/apigateway/ApiGatewayHandlerTest.java
```diff
@@ -66,14 +68,21 @@ class ApiGatewayHandlerTest {
...
     public static final String BOTTOM_EXCEPTION_MESSAGE = "BOTTOM Exception";
-    public static final String SOME_REQUEST_ID = "RequestID:123456";
     public static final int OVERRIDDEN_STATUS_CODE = 418;  //I'm a teapot
     public static final Path EVENT_WITH_UNKNOWN_REQUEST_INFO = Path.of("apiGatewayMessages",
                                                                        "eventWithUnknownRequestInfo.json");
     private static final String PATH = "path1/path2/path3";
     private Context context;
     private Handler handler;
 
+    public static Stream<String> mediaTypeProvider() {
```
brinxmat, line 78: I created the argument provider for a parametrised test, rather than testing individually. I think the tests in this file should be refactored, but this will have to be addressed later.

apigateway/src/test/java/nva/commons/apigateway/ApiGatewayHandlerTest.java
```diff
@@ -111,6 +120,9 @@ public void handleRequestShouldHaveAvailableTheRequestHeaders() throws IOExcepti
                                                               expectedHeaders.get(expectedHeader).textValue()))));
     }
 
+    // TODO: Should return 415 when the Content-type header of request is unsupported (i.e., the one
+    //  describing the content of the body of a post request)
+    // TODO: Should return 406 when Accept header contains unsupported media type
```
brinxmat, line 125: The return code here is simply incorrect, we should fix this asap.

apigateway/src/test/java/nva/commons/apigateway/ApiGatewayHandlerTest.java
```diff
@@ -155,17 +167,18 @@ public void handleRequestReturnsContentTypeJsonOnAcceptWildcard()
         assertThat(response.getHeaders().get(CONTENT_TYPE), is(equalTo(MediaType.JSON_UTF_8.toString())));
     }
 
-    @Test
-    public void handleRequestReturnsContentTypeJsonLdOnAcceptJsonLd()
+    @ParameterizedTest(name = "Should return supported type {0} when it is requested")
+    @MethodSource("mediaTypeProvider")
+    void shouldReturnContentTypeMatchingSupportedMediaTypeWhenSupportedMediaTypeIsRequested(String mediaType)
```
brinxmat, line 172: Refactored the existing test to cover all cases

s3/build.gradle
```diff
@@ -10,7 +10,7 @@ dependencies {
     implementation libs.aws.sdk2.core
     implementation libs.aws.ion
 
-    implementation libs.jackson.databind
+    implementation libs.bundles.jackson
```
brinxmat, line 13: Unfortunately, the code here requires the bundle (the jackson extensions were missing and causing errors in other places), or at least most of it. I'm not happy with including the entire bundle, but I don't see an easy way around it without causing extra problems later. Either way, I don't think it bulks the downstream code particularly since you're using jackson anyway.

apigateway/src/test/java/nva/commons/apigateway/ApiGatewayHandlerTest.java
```diff
@@ -132,7 +144,7 @@ public void handleRequestShouldReturnUnsupportedMediaTypeOnUnsupportedAcceptHead
         "*/*",
         "application/json",
         "application/json; charset=UTF-8",
-        "text/html, application/xhtml+xml, application/xml;q=0.9, image/webp, */*;q=0.8"
+        "text/html, application/xhtml+xml, application/xml;q=0.9, image/webp, */*;q=0.8",
```
torarnet, line 147: why the comma?

torarnet (dismissed):

torarnet (approved):

PR #408, opened 2022-11-22
Add secret writer

StigNorland:


StigNorland:
#NP-18587

secrets/src/test/java/nva/commons/secrets/SecretsWriterTest.java
```diff
@@ -0,0 +1,115 @@
...
+    private Credentials createCredentialsMock() {
+        return new Credentials(JSON_SECRET_VALUE_USERNAME, JSON_SECRET_VALUE_PASSWORD);
+    }
+
+    @Test
+    @DisplayName("Update Secret String successfully")
+    void assertUpdateSecretOK() {
+
+        var putResponse = secretsWriter.updateSecret(SECRET_NAME, SECRET_VALUE);
+        assertEquals(putResponse.name(), SECRET_NAME);
```
brinxmat, line 51: You're using JUnit assertions, so the assertion should be inverted — e.g. asssertX(expected, actual), which does seem weird if you're used to Hamcrest. Just use Hamcrest:
```
        assertThat(putResponse.name(), is(equalTo(SECRET_NAME)));
```

secrets/src/test/java/nva/commons/secrets/SecretsWriterTest.java
```diff
@@ -0,0 +1,115 @@
...
+        var putResponse = secretsWriter.updateSecret(JSON_SECRET_NAME, JSON_SECRET_VALUE);
+        assertEquals(putResponse.name(), JSON_SECRET_NAME);
+    }
+
+    @Test
+    @DisplayName("Update Secret Object successfully")
+    void assertUpdateSecretObjectOK() {
+
+        var putResponse = secretsWriter.updateClassSecret(JSON_SECRET_NAME, credentials);
+        assertEquals(putResponse.name(), JSON_SECRET_NAME);
```
brinxmat, line 67: Ditto Hamcrest

secrets/src/test/java/nva/commons/secrets/SecretsWriterTest.java
```diff
@@ -0,0 +1,115 @@
...
+        credentials = createCredentialsMock();
+    }
+
+    private Credentials createCredentialsMock() {
+        return new Credentials(JSON_SECRET_VALUE_USERNAME, JSON_SECRET_VALUE_PASSWORD);
+    }
+
+    @Test
+    @DisplayName("Update Secret String successfully")
+    void assertUpdateSecretOK() {
```
brinxmat, line 48: I see you've copied the style used elsewhere here, but when I refactor anything in commons, I tend to fix he naming too.

We should prefer BDD style naming, which also means I drop display names.

Delete the display names, rename in pattern:

`shouldReturnPersistedSecretNameResponseWhenSecretIsUpdated`

Actually, what should be tested here? The OK response is what was in the. name…probably we want to validate the response according to expectations beyond the name?

secrets/src/test/java/nva/commons/secrets/SecretsWriterTest.java
```diff
@@ -0,0 +1,115 @@
...
+
+        var putResponse = secretsWriter.updateSecret(SECRET_NAME, SECRET_VALUE);
+        assertEquals(putResponse.name(), SECRET_NAME);
+    }
+
+    @Test
```
brinxmat, line 54: This does the same as the prior test, re-write the first test as a parameterized test with the secret value as the parameter.

secrets/src/test/java/nva/commons/secrets/SecretsWriterTest.java
```diff
@@ -0,0 +1,115 @@
...
+    @DisplayName("Update Secret Object successfully")
+    void assertUpdateSecretObjectOK() {
+
+        var putResponse = secretsWriter.updateClassSecret(JSON_SECRET_NAME, credentials);
+        assertEquals(putResponse.name(), JSON_SECRET_NAME);
+    }
+
+    @Test
+    @DisplayName("Update Secret Logs Error When Wrong Secret Name Is Given")
```
brinxmat, line 71: You can parameterize this test (removing the next test in the process) with a @MethodSource providing null, "xx", "xx", null and null, null

brinxmat: Also, what about empty strings?

secrets/src/main/java/nva/commons/secrets/ErrorWritingSecretException.java
```diff
@@ -0,0 +1,17 @@
...
+import nva.commons.core.JacocoGenerated;
+
+public class ErrorWritingSecretException extends RuntimeException {
+
+    public static final String COULD_NOT_WRITE_SECRET_ERROR = "Could not write secret: ";
+
+    @JacocoGenerated
+    public ErrorWritingSecretException() {
+        super(COULD_NOT_WRITE_SECRET_ERROR);
+    }
```
brinxmat, line 12: I'd remove this

secrets/src/main/java/nva/commons/secrets/SecretsWriter.java
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Updates a secret object (json) in AWS Secrets Manager as tObjectJsonSecretValue.
+     *
+     * @param secretName             the user-friendly id of the secret or the secret ARN
+     * @param tObjectJsonSecretValue the class or interface of the class to be persisted
+     * @param <T>                    the type of the class or interface of the class to be persisted
+     * @return PutSecretValueResponse
+     * @throws ErrorReadingSecretException when any error occurs.
+     */
+    public <T> PutSecretValueResponse updateClassSecret(String secretName, T tObjectJsonSecretValue) {
```
brinxmat, line 67: In many ways, the signature explains the naming — maybe I'm weird thinking that it could be called `updateSecret(String secretName, T secretClassInstance)`

secrets/src/main/java/nva/commons/secrets/SecretsWriter.java
```diff
@@ -0,0 +1,109 @@
...
+    private PutSecretValueResponse updateSecretFromAws(String secretName, String value) {
+        return awsSecretsManager.putSecretValue(
+            PutSecretValueRequest.builder()
+                .secretId(secretName)
+                .secretString(value)
+                .build());
+    }
+
+    private DefaultPrettyPrinter getPrettyPrinterCompact() {
+        DefaultPrettyPrinter p = new DefaultPrettyPrinter();
```
brinxmat, line 92: var(s)

brinxmat (changes requested):

secrets/src/test/java/nva/commons/secrets/SecretsWriterTest.java
StigNorland: fixed

secrets/src/test/java/nva/commons/secrets/SecretsWriterTest.java
StigNorland: I have dyslexia and that is reeeeeally hard to read for me.
which mean I can read each word, but I have no idea what it says afterwards, if it is a sentence, no problem.

StigNorland: on the flip side, my analytical abilities are on pair with **_really_** smart ppl, albeit much slower. (consequence of dyslexia)

secrets/src/main/java/nva/commons/secrets/ErrorWritingSecretException.java
StigNorland: I removed the other constructor, as  I think the intention was to avoid leaking secrets in errorlogs

brinxmat (approved):

secrets/src/test/java/nva/commons/secrets/SecretsWriterTest.java
```diff
@@ -22,24 +25,29 @@
...
+    private static final String SECRET_VAULT_ID = "test_app_secret_vault_id";
     private final SecretsWriter secretsWriter;
     private final Credentials credentials;
+    private final ObjectMapper objectMapper =
+        new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
 
     private static Stream<Arguments> invalidArgumentProvider() {
         return Stream.of(
-            Arguments.of(SECRET_NAME, null),
+      //      Arguments.of(SECRET_NAME, null),
```
brinxmat, line 50: Should this be deleted?

secrets/src/main/java/nva/commons/secrets/SecretsWriter.java
```diff
@@ -80,18 +89,26 @@ public static SecretsManagerClient defaultSecretsManagerClient() {
...
+                   .map(secretJson -> {
+
+                       ((ObjectNode) secretJson).put(secretKey, secretValue);
+                       return updateSecretJsonToAws(secretValueResponse.name(), secretJson);
+
+                   }).orElseThrow(this::logErrorAndThrowException);
     }
 
-    private <T> PutSecretValueResponse updateSecretJsonToAws(String secretName, T node) {
+    private <T> PutSecretValueResponse updateSecretJsonToAws(String secretVaultId, T node) {
```
brinxmat, line 108: Can you update the variable `secretVaultId` name to `secretName` (so it is called this consistently in the API both from AWS and NVA)

brinxmat (changes requested):

StigNorland:
Yes, my mistake.

/Stig

brinxmat (approved):

PR #411, opened 2022-12-06
unmark apigatewayv2 from being a submodule

axthosarouris:
ApiGatewayV2 was an attempt to make the ApiGateway handlers faster by using jackson-jr instead of standard jackson  and sdk2. Sdk2 is now used by ApiGateway and jackson-jr did not seem to offer a speed advantage. ApiGatewayv2 was unsuccessful and its existence creates confusion. 

In this PR I am only removing it from the settings. In subsequent PR we can delete the folder


sondrev (approved):
Sounds good. I cant see apigateway2 used from any other project

PR #413, opened 2022-12-22
Update aws sdk

axthosarouris:


buildSrc/src/main/groovy/nvacommons.java-conventions.gradle
```diff
@@ -9,7 +9,7 @@ plugins {
 }
 
 group 'com.github.bibsysdev'
-version = '1.26.1'
+version = '1.27.0'
```
axthosarouris, line 12: going to 1.27.0 because updating nva-commons will mean that the services that use commons will also need to update their aws sdk versions.

config/checkstyle/checkstyle.xml
```diff
@@ -286,7 +286,7 @@
     </module>
 
     <module name="JavadocMethod">
-      <property name="scope" value="public"/>
+      <property name="accessModifiers" value="public"/>
```
axthosarouris, line 289: Checkstyle was complaining that this field does not exist: https://checkstyle.sourceforge.io/config_javadoc.html#JavadocMethod

gradle/libs.versions.toml
```diff
@@ -2,8 +2,8 @@
 zalandoProblem = { strictly = '0.27.1' }
 lambdaLog4j = { strictly = '1.5.1' }
 guava = { strictly = '31.1-jre' }
-awsSdk = { strictly = '1.12.220' }
-awsSdk2 = { strictly = '2.17.191' }
+awsSdk = { strictly = '1.12.369' }
```
axthosarouris, line 5: At the time of PR, these are the latest versions

config/checkstyle/checkstyle.xml
axthosarouris: Failing in my machine because of a more recent gradle version. Reverted back to "scope"

brinxmat (approved):

PR #418, opened 2023-02-07
moving PagedSearchResult<T> to nva-commons

ketilaa:
I am yet again to implement a search endpoint with pagination, and want to move it to nva-commons from nav-verified-funding-sources.

Current implementation has the following features:

- fixed query parameters `query` and `size` for pagination
- currently no explicit support for real searches in generation if `id` (self URI) as search term won't be included unless it is added to `baseUri`
- personally, I find it a bit confusing that we use name `size` for both the query parameter and in the response for the total size of the entire result set. `totalSize` is a better name in the response in my opinion.

I know this is a bit different from the way we do in cristin-service where `page` and `results` are used as query parameters.


pagination/src/main/java/no/unit/nva/commons/pagination/PagedSearchResult.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public static final String OFFSET_QUERY_PARAM_NAME = "offset";
+    public static final String SIZE_QUERY_PARAM_NAME = "size";
+
+    @JsonProperty(CONTEXT_FIELD_NAME)
+    private final URI context;
+    @JsonProperty(ID_FIELD_NAME)
+    private final URI id;
+    @JsonProperty(SIZE_FIELD_NAME)
+    private final int totalSize;
```
brinxmat, line 38: perhaps "total hits" since this is what is counted?

buildSrc/src/main/groovy/nvacommons.java-conventions.gradle
```diff
@@ -9,7 +9,7 @@ plugins {
 }
 
 group 'com.github.bibsysdev'
-version = '1.27.3'
+version = '1.28.0'
```
brinxmat, line 12: Thinking about this, does this not belong in its own search-commons module?

torarnet (dismissed):

pagination/src/main/java/no/unit/nva/commons/pagination/PagedSearchResult.java
monaullah: Done

torarnet (approved):

pagination/src/main/java/no/unit/nva/commons/pagination/PagedSearchResult.java
```diff
@@ -0,0 +1,148 @@
...
+                             int queryOffset,
+                             int querySize,
+                             int totalHits,
+                             List<T> hits,
+                             Map<String, String> queryParameters) {
+
+        URI selfUri = generateSelfUri(baseUri, queryOffset, querySize, queryParameters);
+        URI nextResults = calculateNextResults(queryOffset, querySize, totalHits, hits.size(), baseUri,
+                                               queryParameters);
+        URI previousResults = calculatePreviousResults(queryOffset, querySize, baseUri, queryParameters);
```
brinxmat, line 82: var

pagination/src/main/java/no/unit/nva/commons/pagination/PagedSearchResult.java
```diff
@@ -0,0 +1,148 @@
...
+    private final int totalHits;
+    @JsonProperty(NEXT_RESULTS_FIELD_NAME)
+    private final URI nextResults;
+    @JsonProperty(PREVIOUS_RESULTS_FIELD_NAME)
+    private final URI previousResults;
+    @JsonProperty(HITS_FIELD_NAME)
+    private final List<T> hits;
+
+    @JsonCreator
+    public PagedSearchResult(@JsonProperty(CONTEXT_FIELD_NAME) URI context,
```
brinxmat, line 48: A comment that is not directly actionable, but is worth thinking about: bizarrely, the context is always the same for this object. We have made it configurable because we don't know what this URI is, I guess. Is it unwise to fix this URI in this object because the object is "generic", or because we don't reliably serve the URI?

pagination/src/main/java/no/unit/nva/commons/pagination/PagedSearchResult.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    public List<T> getHits() {
+        return hits;
+    }
+
+    private static URI calculateNextResults(int queryOffset,
+                                            int querySize,
+                                            int totalHits,
+                                            int noHits,
```
brinxmat, line 118: numberOfHits — I got confused by the naming

pagination/src/main/java/no/unit/nva/commons/pagination/PagedSearchResult.java
```diff
@@ -0,0 +1,148 @@
...
+                                            URI baseUri,
+                                            Map<String, String> queryParams) {
+        if ((queryOffset + noHits) < totalHits) {
```
brinxmat, line 121: Is it less than or equal to, perhaps?

brinxmat: I'd probably test it

brinxmat (changes requested):

pagination/src/main/java/no/unit/nva/commons/pagination/PagedSearchResult.java
monaullah: Done

pagination/src/main/java/no/unit/nva/commons/pagination/PagedSearchResult.java
monaullah: Agreed

pagination/src/main/java/no/unit/nva/commons/pagination/PagedSearchResult.java
monaullah: Good point. Made context constant (https://bibsysdev.github.io/src/search/paginated-search-result.json), will define asap.

pagination/src/test/java/no/unit/nva/commons/pagination/PaginatedSearchResultTest.java
```diff
@@ -0,0 +1,122 @@
...
+import java.util.Set;
+import nva.commons.core.paths.UriWrapper;
+import org.junit.jupiter.api.Test;
+
+class PaginatedSearchResultTest {
+
+    public static final String QUERY_PARAM_FIELD_NAME = "key";
+    public static final String QUERY_PARAM_FIELD_VALUE = "value";
+    private static final URI BASE_URI = URI.create("https://localhost");
+
```
brinxmat, line 25: Add a test for nonsense values:

```
    @ParameterizedTest
    @CsvSource({"10,0", "-1,10", "10,-1"})
    void shouldThrowOnNonsenseRequest(int offset, int size) {
        assertThrows(Exception.class, () -> PaginatedSearchResult.create(BASE_URI, offset, size, 50,
                                                                 Collections.emptyList(), Collections.emptyMap()));
    }
```

brinxmat: Interestingly, I'm not sure what the exception should be. The server should return something that is in the 4XX range, but there are a few candidates. 422 with a message would probably be best, but I have seen 400 (which is wrong for good reasons: the request _is_ valid syntactically) and 404 suggested. I think 422 _is_ better, see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422, but it is also a bit uncommon.

pagination/src/test/java/no/unit/nva/commons/pagination/PaginatedSearchResultTest.java
```diff
@@ -0,0 +1,122 @@
...
+    @Test
+    void shouldPopulateContextIdTotalHitsAndHitsAlways() {
+        var result = PaginatedSearchResult.create(BASE_URI, 0, 5, 0, Collections.emptyList());
+
+        assertThat(result, doesNotHaveEmptyValuesIgnoringFields(Set.of("nextResults", "previousResults", "hits")));
+        assertThat(result.getHits(), emptyIterable());
+    }
+
+    @Test
```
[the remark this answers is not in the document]
brinxmat: Another option is to say that this is also an exception.

pagination/src/main/java/no/unit/nva/commons/pagination/PaginatedSearchResult.java
```diff
@@ -0,0 +1,147 @@
...
+    private final int totalHits;
+    @JsonProperty(NEXT_RESULTS_FIELD_NAME)
+    private final URI nextResults;
+    @JsonProperty(PREVIOUS_RESULTS_FIELD_NAME)
+    private final URI previousResults;
+    @JsonProperty(HITS_FIELD_NAME)
+    private final List<T> hits;
+
+    @JsonCreator
+    public PaginatedSearchResult(@JsonProperty(ID_FIELD_NAME) URI id,
```
brinxmat, line 51: Since we want to force use of the factory methods "create", hide the constructor.

brinxmat (changes requested):

pagination/src/main/java/no/unit/nva/commons/pagination/PagedSearchResult.java
brinxmat: I tried to address my confusion — for some reason I was expecting a different order in the code. I am also good at maths (ahem) :D

pagination/src/test/java/no/unit/nva/commons/pagination/PaginatedSearchResultTest.java
monaullah: See https://github.com/BIBSYSDEV/nva-commons/pull/428

brinxmat (approved):

PR #420, opened 2023-02-28
Fix bug in S3Driver: Depends on PR #421

axthosarouris:
S3Driver uses the ListObjectsRequest that does not seem to work properly. This PR makes the S3Driver use the ListObjetsV2Request and the respective s3 method.

The ListObjectsV2Request obfuscates the nest starting point so it is safer as well.


nvatestutils/src/main/java/no/unit/nva/stubs/FakeS3Client.java
```diff
@@ -111,17 +98,29 @@ public ListObjectsResponse listObjects(ListObjectsRequest listObjectsRequest) {
...
                 .isTruncated(nonNull(nextStartListingPoint)).build();
     }
 
-    private String calculateNestStartListingPoint(List<String> fileKeys,
-                                                  int excludedEndIndex) {
-        return excludedEndIndex >= fileKeys.size()
-                ? null
-                : fileKeys.get(excludedEndIndex-1);
+    @Override
+    public ListObjectsV2Response listObjectsV2(ListObjectsV2Request v2Request){
```
axthosarouris, line 106: this is the new method.

nvatestutils/src/main/java/no/unit/nva/stubs/FakeS3Client.java
```diff
@@ -141,6 +140,28 @@ public String serviceName() {
     public void close() {
 
     }
+    private static ByteBuffer readFileFromResources(String filename) {
```
axthosarouris, line 143: moved due to auto formatting.

brinxmat (approved):

PR #430, opened 2023-05-05
NP-44689-import-candidates-access-rights

anetteOlli:


ketilaa (dismissed):
Note that we need a role to be able to give this access right to someone in the future.

anetteOlli:
Rollene lages vel i identity service? Denne må merges før det kan gjøres.

brinxmat (approved):

PR #433, opened 2023-05-09
UnixPath - get path element by index from end

monaullah:


core/src/main/java/nva/commons/core/paths/UnixPath.java
```diff
@@ -102,6 +102,10 @@ public String getLastPathElement() {
         return path.get(lastPathElementIndex());
     }
 
+    public String getSecondToLastPathElement() {
```
brinxmat, line 105: Probably just:

```
    public String getPathElementByIndex(int index) {
        return path.get(index);
    }
    
    public String getPathElementByIndexFromEnd(int index) {
      return path.get(path.size() - index);
   }
```

brinxmat: Naming…yep, I know.

brinxmat (changes requested):

brinxmat (changes requested):

core/src/test/java/nva/commons/core/paths/UnixPathTest.java
```diff
@@ -186,22 +187,32 @@ void shouldReturnTheLastElementOfaUnixPath(String inputPath, String expectedFile
...
+
+    @Test
+    void shouldReturnPathElementByIndexFromEndOfaUnixPath() {
+        var pathElements = List.of("first", "second", "third", "fourth", "fifth").toArray(new String[0]);
+        var unixPath = UnixPath.of(pathElements);
+
+        for (int i = 0; i < pathElements.length; i++) {
+            assertThat(unixPath.getPathElementByIndexFromEnd(pathElements.length - i), is(equalTo(pathElements[i])));
+        }
+    }
```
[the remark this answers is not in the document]
monaullah: Done, with an adjustment: `assertThat(unixPath.getPathElementByIndexFromEnd(value + 1), is(equalTo(expected)));`

brinxmat (approved):

brinxmat (approved):

PR #436, opened 2023-05-15
firehose-client

axthosarouris:
A Kinesis Firehose collects the records that are pushed to it and stores them in an S3 bucket.
  The important feature is that it collects multiple records in ZIP files and organizes the ZIP files by timestamp:
  s3://some-bucket/some-folder/2023/05/08/18/some-prefix-2023-05-08-18-01-17-lfksdjflskd-srdfsdef-sefsdf-sdfsd
  -sgdfsdfsdfs.gz

This a stub for Firehose client for testing

### DIFF
--- a/nvatestutils/src/test/java/no/unit/nva/stubs/FakeFirehoseClientTest.java
+++ b/nvatestutils/src/test/java/no/unit/nva/stubs/FakeFirehoseClientTest.java
@@ -37,7 +37,7 @@ public class FakeFirehoseClientTest {
     void shouldReturnAllRecordsWhenPushingABatchOfRecords() {
         var expectedRecords = List.of(randomRecord(), randomRecord());
 
-        PutRecordBatchRequest request = PutRecordBatchRequest.builder()
+        var request = PutRecordBatchRequest.builder()
                                             .records(expectedRecords)
                                             .build();
         client.putRecordBatch(request);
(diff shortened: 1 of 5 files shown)


nvatestutils/src/test/java/no/unit/nva/stubs/FakeFirehoseClientTest.java
```diff
@@ -0,0 +1,55 @@
...
+        client.putRecord(request);
+        var actualRecord = client.getRecords().stream().collect(SingletonCollector.collect());
+        assertThat(actualRecord, is(equalTo(expectedRecord)));
+    }
+
+    @Test
+    void shouldReturnAllRecordsWhenPushingABatchOfRecords() {
+        var expectedRecords = List.of(randomRecord(), randomRecord());
+
+        PutRecordBatchRequest request = PutRecordBatchRequest.builder()
```
brinxmat, line 40: var

brinxmat (changes requested):

anetteOlli (dismissed):
Side note:
Det ser ut til at det har kommet på plass noe bibliotek for testing med aws resurser:
https://madhead.github.io/aws-junit5/asciidoc/

axthosarouris:
> Side note: Det ser ut til at det har kommet på plass noe bibliotek for testing med aws resurser: https://madhead.github.io/aws-junit5/asciidoc/

Nice! I think this is to be used in combination with localstack which could be very interesting. It is certainly worth a look and a better solution than the custom implemented testing clients

brinxmat (approved):

PR #438, opened 2023-06-01
first version of dlq handler

axthosarouris:
DLQ handler for using in Lambda destinations. Look at the comments for more information.

The PR is large due to the example Cloudformation templates.


dlq/src/main/java/no/unit/nva/commons/dlq/DlqHandler.java
```diff
@@ -0,0 +1,84 @@
...
+    private final FirehoseClient firehoseClient;
+
+    public DlqHandler(FirehoseClient firehoseClient) {
+        this.firehoseClient = firehoseClient;
+    }
+
+    @Override
+    public Void handleRequest(SQSEvent input, Context context) {
+
+        var failedEvents = input.getRecords().stream()
```
axthosarouris, line 53: each DLQ event might contain multiple records

dlq/src/test/resources/examples/dlq-stack.yaml
```diff
@@ -0,0 +1,189 @@
+AWSTemplateFormatVersion: "2010-09-09"
```
axthosarouris, line 1: Example DLQ stack that contains all resources and exposes only the DLQ ARN and the policy that allows lambdas to write to the DLQ. 
It can be deployed independently (i.e. one DLQ stack for the whole account) or as a nested stack (i.e. one nested DLQ stack for each deployed stack)

dlq/src/test/resources/examples/use_dlq_stack.yaml
```diff
@@ -0,0 +1,72 @@
+AWSTemplateFormatVersion: "2010-09-09"
```
axthosarouris, line 1: Example of using the DLQ stack as nested stack

dlq/src/main/java/no/unit/nva/commons/dlq/DlqHandler.java
axthosarouris: Probably yes. This handler however makes no assumptions for the SQS queue.6 If the queue has batch size 1 then each message will be handled separately.

anetteOlli: I would prefer it if the failure of individual messages didn't cause the lambda to throw exception. Because if the batch always comes in the same order, we have a problem. Lets say message number 3 of 10 fails every time. Then messages 4 to 10 will never be processed. 
Isn't there supposed to be an object the lambda can return with list of unsuccessful messages?

But I think leaving it as it is now is okay, and then observe if we need to change it later is okay.

anetteOlli (approved):

dlq/src/main/java/no/unit/nva/commons/dlq/DlqHandler.java
```diff
@@ -0,0 +1,89 @@
...
+ * See more detailed example in the test resources.
+ */
+public class DlqHandler implements RequestHandler<SQSEvent, Void> {
+
+    public static final Environment ENVIRONMENT = new Environment();
+    public static final String DELIVERY_STREAM_NAME = ENVIRONMENT.readEnv("DELIVERY_STREAM_NAME");
+    public static final int NUMBER_OF_GROUPS = 10;
+    private final FirehoseClient firehoseClient;
+
+    public DlqHandler(FirehoseClient firehoseClient) {
```
brinxmat, line 51: There is no default constructor here

dlq/src/main/java/no/unit/nva/commons/dlq/DlqHandler.java
```diff
@@ -0,0 +1,89 @@
...
+        this.firehoseClient = firehoseClient;
+    }
+
+    @Override
+    public Void handleRequest(SQSEvent input, Context context) {
```
brinxmat, line 56: I suspect moving the logic to some class makes the code easier to comprehend.

```
package no.unit.nva.commons.dlq;

import com.google.common.collect.Lists;
import java.nio.charset.StandardCharsets;
import java.util.Collection;
import java.util.List;
import java.util.stream.Collectors;
import software.amazon.awssdk.core.SdkBytes;
import software.amazon.awssdk.services.firehose.FirehoseClient;
import software.amazon.awssdk.services.firehose.model.PutRecordBatchRequest;
import software.amazon.awssdk.services.firehose.model.Record;

public class FireHoseService {
    public static final int NUMBER_OF_GROUPS = 10;
    private final FirehoseClient client;
    private final String streamName;

    public FireHoseService(FirehoseClient client, String streamName) {
        this.client = client;
        this.streamName = streamName;
    }

    // API-wise, you might be tempted to return something other than Void, which can be used by the receiver.
    public Void persist(List<String> messages) {
        splitToSubLists(messages).forEach(this::pushToFirehose);
        return null;
    }

    private static Record createFirehoseRecord(String failedEvent) {
        return Record.builder().data(SdkBytes.fromString(failedEvent, StandardCharsets.UTF_8)).build();
    }

    private static Collection<List<String>> splitToSubLists(List<String> failedEvents) {
        return Lists.partition(failedEvents, NUMBER_OF_GROUPS);
    }

    private PutRecordBatchRequest assemblePutBatchRequest(List<Record> records) {
        return PutRecordBatchRequest.builder()
                   .records(records)
                   .deliveryStreamName(streamName)
                   .build();
    }

    private void pushToFirehose(List<String> group) {
        var records = group.stream()
                          .map(FireHoseService::createFirehoseRecord)
                          .collect(Collectors.toList());
        var request = assemblePutBatchRequest(records);
        client.putRecordBatch(request);
    }

```

brinxmat: Also, it looks like the NUMBER:OF_GROUPS is a default value, which needs configuring.

dlq/src/test/java/no/unit/nva/commons/dlq/DlqHandlerTest.java
```diff
@@ -0,0 +1,86 @@
...
+    @BeforeEach
+    public void init() {
+        this.firehoseClient = new FakeFirehoseClient();
+        this.handler = new DlqHandler(firehoseClient);
+    }
+
+    @ParameterizedTest(name = "number of messages:{1}")
+    @MethodSource("eventProvider")
+    void shouldPushFailedEventsToFirehoseForAutomaticallyStoringThemInS3AndOrganizeThemByTime(
+        Set<JsonNode> failedEvents, int numberOfEvents) {
```
brinxmat, line 48: `numberOfEvents` seems unused except in the test name, perhaps intentional, but a bit weird. If you want to affect changes in the name, probably (NB: you'll need to remove the name from `@ParameterizedTest`):

```
    public static Stream<Arguments> eventProvider() {
        return IntStream.of(1, 2, 10, 245).boxed()
                   .map(numberOfMessages ->
                        Arguments.of(Named.of(numberOfMessages +  " messages", randomJsons(numberOfMessages))));
    }
```

dlq/src/test/java/no/unit/nva/commons/dlq/DlqHandlerTest.java
```diff
@@ -0,0 +1,86 @@
...
+
+        var dlqEvent = createDlqEventContainingFailedEvents(failedEvents);
+        handler.handleRequest(dlqEvent, EMPTY_CONTEXT);
+        var pushedRecords = firehoseClient.getRecords().stream()
+                                .map(record -> record.data().asUtf8String())
+                                .map(attempt(OBJECT_MAPPER::readTree))
+                                .map(Try::orElseThrow)
+                                .collect(Collectors.toSet());
```
brinxmat, line 56: Tempted to say that this is a suitably named method in FakeFirehoseClient

dlq/src/test/resources/examples/use_dlq_stack.yaml
```diff
@@ -0,0 +1,72 @@
...
+  - AWS::Serverless-2016-10-31
+  - AWS::LanguageExtensions
+Description: Aria Events Validation
+
+Globals:
+  Function:
+    MemorySize: 1792
+    Tracing: Active
+    Timeout: 60
+    Runtime: java11
```
brinxmat, line 12: Still no Java17 for us :D

brinxmat (changes requested):

dlq/src/main/java/no/unit/nva/commons/dlq/DlqHandler.java
axthosarouris: It is intentional. added documentation and a SampleDlqHandler in the tests

dlq/src/main/java/no/unit/nva/commons/dlq/DlqHandler.java
axthosarouris: Will do in next PR.

dlq/src/main/java/no/unit/nva/commons/dlq/DlqHandler.java
axthosarouris: I will look into it.

> Isn't there supposed to be an object the lambda can return with list of unsuccessful messages?
I think this would be a DLQ for the DLQ :)

Bear in mind that we are talking about persisting a failed event, so a failure here would be failure to persist an already failed event. I can see the reason why we would like to persist as many failed events as possible, but at the same time, failing to do the one and only thing this handler is supposed to be doing should be an exception. Perhaps the solution would be to persist as many as and then throw an Exception. 
That being said, given that the handler is pushing Strings down to a firehose, failing to do so for one string would most likely mean that it would fail to do so for any subsequent string.

dlq/src/main/java/no/unit/nva/commons/dlq/PushToFirehoseService.java
```diff
@@ -0,0 +1,48 @@
...
+import nva.commons.core.Environment;
+import software.amazon.awssdk.core.SdkBytes;
+import software.amazon.awssdk.services.firehose.FirehoseClient;
+import software.amazon.awssdk.services.firehose.model.PutRecordBatchRequest;
+import software.amazon.awssdk.services.firehose.model.Record;
+
+public class PushToFirehoseService implements FailedEventHandlingService {
+
+    public static final Environment ENVIRONMENT = new Environment();
+    public static final String DELIVERY_STREAM_NAME = ENVIRONMENT.readEnv("DELIVERY_STREAM_NAME");
```
brinxmat, line 16: Can we pull these up to the handler? I feel like this is quite specific config, along with the batch size. I guess a config object can be injected to allow this.

dlq/src/test/java/no/unit/nva/commons/dlq/DlqHandlerTest.java
brinxmat: I still think this :)

brinxmat (changes requested):

axthosarouris:
I have removed all configuration and environment variables. There is no reason for them at this level, since the class needs to be subclassed to be used. The subclass will have all the configuration. Removed also the default methods because if we do not have any configuration, then we are just passing the same arguments ( fireshoseClient and firehose name) around.

brinxmat (approved):

PR #439, opened 2023-06-01
NP-44689-import-candidates-access-rights, added access right

anetteOlli:


apigateway/src/main/java/nva/commons/apigateway/AccessRight.java
```diff
@@ -18,6 +18,7 @@ public enum AccessRight {
     READ_DOI_REQUEST,
     APPROVE_PUBLISH_REQUEST,
     EDIT_OWN_INSTITUTION_RESOURCES,
+    EDIT_ALL_RESOURCES,
```
StigNorland, line 21: Is this an admin AccessRight ? 
How about THESIS?

StigNorland (approved):

ketilaa (approved):

apigateway/src/main/java/nva/commons/apigateway/AccessRight.java
StigNorland: Name was fine with me, and I don't have a suggestion for a better one.

brinxmat (approved):

PR #465, opened 2023-11-24
adding custom object to ApigGatewayException

truhacevkir:
This makes it possible to return problem+json containing some object, not only error detail. 
This will be used to return tombstone to frontend:

```
{
      statusCode: ...
      detail: ...
      resource: {
              .....
              .....
              .....
      }
}
```


brinxmat (changes requested):
Fix the Codacy issues

PR #472, opened 2023-12-20
NP-45696-doi-without-protocol

anetteOlli:


doi/src/test/java/nva/commons/doi/DoiConverterTest.java
```diff
@@ -110,6 +115,14 @@ public void toUriThrowsExceptionWhenInputIsAnInvalidDoiString() {
         assertThat(appender.getMessages(), containsString(input));
     }
 
+    @ParameterizedTest(name = "should add https for doi missing protocol:{0}")
```
brinxmat, line 118: `protocol` -> `scheme`

brinxmat: As elsewhere :)

brinxmat (approved):

PR #494, opened 2024-05-16
FakeS3Client listObjectsV2 fix startAfter

monaullah:


nvatestutils/src/test/java/no/unit/nva/stubs/FakeS3ClientTest.java
```diff
@@ -120,6 +124,22 @@ void shouldReturnNextListingStartPointWhenReturningAPageOfResultsWithVersion2Req
         assertThat(result.nextContinuationToken(), is(equalTo(expectedFile1.toString())));
     }
 
+    @Test
+    void shouldReturnNextListingStartPointWhenReturningAPageOfResultsWithVersion2Request() {
+        var s3Client = new FakeS3Client();
+        var bucket = randomString();
+        var expectedFile1 = insertRandomFileToS3(s3Client, bucket);
+        var otherFile = insertRandomFileToS3(s3Client, bucket);
```
axthosarouris, line 132: I think the "startAfter" has nothing to do with pagination. From
 https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/model/ListObjectsV2Response.html
```
NextContinuationToken is sent when isTruncated is true, which means there are more keys in the bucket that can be listed.

If StartAfter was sent with the request, it is included in the response.
```
I think pagination in s3 listing  cannot be done with startAfter because this key will always be a copy of the "startAfter" field of the request. 

I remember that was tricky part. I do not think the current code is obfuscating the keys since it is a fake client. However I would suggest to create a bucket with many files and try to use both startAfter and continuationToken and see which one works.

axthosarouris: I would also recommend to look at http://docs.getmoto.org/en/latest/docs/server_mode.html. This *might* be a better solution than the current approach. Having said that, I am using the same approach :)

monaullah: Even though "startAfter" in the response is a copy of the "startAfter" field of the request, I think it can be used for pagination if the client holds track of the last key in a `ListObjectsV2Response` and uses this as `startAfter` in the next request. I have tested this with expected results with a real client.

From [docs](https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/s3/model/ListObjectsV2Request.html#withStartAfter-java.lang.String-): _Sets the optional parameter indicating where you want Amazon S3 to start the object listing from. This can be any key in the bucket._

Have you tested "continuationToken" for pagination with the real client @axthosarouris?

monaullah: I tried testing with a real client and `response.nextContinuationToken` as `continuationToken` in next request and it seems to work just as fine, so I'm moving on with this. Closing this draft

Thanks for the input:)

PR #497, opened 2024-06-02
ApiS3PresignerGatewayHandler

monaullah:
`ApiS3PresignerGatewayHandler`: To handle cases where generation of export data takes too long time (longer than api gateway timeout). Handler that generates and returns a pre-signed s3 url. Abstract method `generateAndWriteDataToS3` lets handler delegate creation of data to different handler.

Similar to `ApiS3GatewayHandler`, but in this handler data creation can be delegated.

### DIFF
--- a/apigateway/src/main/java/nva/commons/apigateway/ApiS3PresignerGatewayHandler.java
+++ b/apigateway/src/main/java/nva/commons/apigateway/ApiS3PresignerGatewayHandler.java
@@ -2,6 +2,7 @@ package nva.commons.apigateway;
 
 import com.amazonaws.services.lambda.runtime.Context;
 import java.net.HttpURLConnection;
+import java.net.URL;
 import java.time.Duration;
 import java.util.Map;
 import nva.commons.apigateway.exceptions.BadRequestException;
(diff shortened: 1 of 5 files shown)


apigateway/src/main/java/nva/commons/apigateway/ApiS3PresignerGatewayHandler.java
```diff
@@ -0,0 +1,67 @@
...
+import nva.commons.core.JacocoGenerated;
+import software.amazon.awssdk.services.s3.model.GetObjectRequest;
+import software.amazon.awssdk.services.s3.presigner.S3Presigner;
+import software.amazon.awssdk.services.s3.presigner.model.GetObjectPresignRequest;
+import software.amazon.awssdk.services.s3.presigner.model.PresignedGetObjectRequest;
+
+public abstract class ApiS3PresignerGatewayHandler<I> extends ApiGatewayHandler<I, Void> {
+
+    public static final String LOCATION = "Location";
+    public static final Duration SIGN_DURATION = Duration.ofMinutes(60);
```
torbjokv, line 18: Maybe we want this to be configurable? was thinking about cases where we want it longer, but shorter is maybe more likely

sondrev (approved):

sondrev (approved):

PR #498, opened 2024-06-03
RestRequestHandler, validateRequest

monaullah:
Initially needed to validate access rights in `ApiS3PresignerGatewayHandler`, but I think it makes more sense to have such a general method in `RestRequestHandler`. Any thoughts?


sondrev (dismissed):
No objections, but please wait to receive feedback from other developers as well

ketilaa (approved):

ketilaa:
The way you introduce this "feature" to the `ApiGatewayHandler` does not encourage future use of it. An option could be to not add the default implementation to force all implementations to migrate when they update `commons`.

brinxmat (approved):

PR #500, opened 2024-06-21
Np 47019 IdentityServiceClient, getUser

monaullah:
New method for fetching user from identity service. Will be used in nvi to fetch users `viewingScope` for task NP-47019.


clients/src/main/java/no/unit/nva/clients/IdentityServiceClient.java
```diff
@@ -90,7 +72,59 @@ public GetExternalClientResponse getExternalClientByToken(String bearerToken)
                    .orElseThrow(this::handleFailure);
     }
 
-    private NotFoundException handleFailure(Failure<GetExternalClientResponse> responseFailure) {
+    public GetUserResponse getUser(String userName) throws NotFoundException {
```
monaullah, line 75: New method

sondrev (approved):

PR #502, opened 2024-07-02
restrict-allow-origin

anetteOlli:
Only one domain can be returned when setting the Access-allow-origin-header.

If we have multiple supported origins, we need to echo the request.header.origin back to the client with the Vary header also set after checking a whitelist.

We probably going to need to be able to set a list of allowed origins, because there are a multiple of frontends that are served from the same backend. Ex. frontenders running their code locally, and having a temporary branch deployed for testers.

This PR will also be a nightmare for our frontenders: https://stackoverflow.com/questions/10883211/why-does-my-http-localhost-cors-origin-not-work

### DIFF
--- a/apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandler.java
+++ b/apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandler.java
@@ -300,7 +300,7 @@ public abstract class ApiGatewayHandler<I, O> extends RestRequestHandler<I, O> {
         headers.put(CONTENT_TYPE, MediaTypes.APPLICATION_PROBLEM_JSON.toString());
         headers.put(X_CONTENT_TYPE_OPTIONS, "nosniff");
         headers.put(STRICT_TRANSPORT_SECURITY, "max-age=63072000; includeSubDomains; preload");
-        headers.put(VARY, "Origin");
+        headers.put(VARY, "Origin, Accept");
         return headers;
     }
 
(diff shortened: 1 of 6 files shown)


apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandler.java
```diff
@@ -281,6 +310,7 @@ private Map<String, String> defaultHeaders(RequestInfo requestInfo) throws Unsup
         headers.put(CONTENT_TYPE, getDefaultResponseContentTypeHeaderValue(requestInfo).toString());
         headers.put(X_CONTENT_TYPE_OPTIONS, "nosniff");
         headers.put(STRICT_TRANSPORT_SECURITY, "max-age=63072000; includeSubDomains; preload");
+        headers.put(VARY, "Origin");
```
anetteOlli, line 313: NB!

vi kan ha behov for å sette Vary headeren under content-negotiation.

apigateway/src/main/java/nva/commons/apigateway/RestRequestHandler.java
```diff
@@ -190,6 +195,8 @@ protected void handleExpectedException(Context context, I inputObject, ApiGatewa
 
     protected void init(OutputStream outputStream, Context context) {
         this.outputStream = outputStream;
+        this.allowedOrigin = ALLOW_ALL_ORIGINS; //anti-pattern: this will be overwritten later on, but in the event
+        // that we get an exception before we get there, there is this.
```
anetteOlli, line 199: Dette er stygt, men kom ikke på en bedre måte å gjøre dette. AllowedOrigin må være satt selv om vi får en exception mens vi parser inputstreamen og dermed ikke klarer å lese ut origin headeren fra requesten.

apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandler.java
brinxmat: Ikke for å nevne de andre tingene vi har glemt å sette her  :D

apigateway/src/main/java/nva/commons/apigateway/RestRequestHandler.java
brinxmat: Can it be explicitly null?

brinxmat (approved):

PR #503, opened 2024-07-04
limit-allow-origin-bug

anetteOlli:
Tydeligvis får vi origin headeren lower case fra ApiGateway.

image


apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandler.java
```diff
@@ -83,7 +84,7 @@ private String readAllowedOrigin(RequestInfo requestInfo) {
         if (originsList.contains(ALL_ORIGINS_ALLOWED)) {
             return ALL_ORIGINS_ALLOWED;
         }
-        var requestOrigin = requestInfo.getHeader("Origin");
```
ketilaa, line 86: HTTP header names should be case insensitive, so this is a weakness in our implementation in RequestInfo I would say.

ketilaa (dismissed):

apigateway/src/main/java/nva/commons/apigateway/ApiGatewayHandler.java
anetteOlli: Good point. Fixed the actual issue.

ketilaa (approved):

PR #508, opened 2024-08-26
Fix bug that occurs when a URI with encoded query params is passed to…

brinxmat:
Fix bug that occurs when a URI with encoded query params is passed to UriWrapper and additional query params are added.

Testing multiple strange configurations since one never knows.

### DIFF
--- a/core/src/main/java/nva/commons/core/paths/UriWrapper.java
+++ b/core/src/main/java/nva/commons/core/paths/UriWrapper.java
@@ -1,13 +1,21 @@
 package nva.commons.core.paths;
 
 import static java.util.Objects.isNull;
+import static java.util.Objects.nonNull;
 import static nva.commons.core.attempt.Try.attempt;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URLDecoder;
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import nva.commons.core.JacocoGenerated;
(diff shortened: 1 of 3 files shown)


core/src/test/java/nva/commons/core/paths/UriWrapperTest.java
```diff
@@ -22,6 +26,29 @@ class UriWrapperTest {
     public static final int MAX_PORT_NUMBER = 65535;
     private static final String ROOT = "/";
 
+    public static Stream<Named<URI>> queryParameterProvider() {
+        return Stream.of(
+            Named.of("URI with unencoded query part", URI.create("https://example.org?q=høhë")),
+            Named.of("URI with encoded query part", URI.create("https://example.org?q=h%C3%B8h%C3%AB"))
```
monaullah, line 32: Add test for uri with several existing query params?
Example `https://example.org?q=høhë&param2=åpen`

core/src/main/java/nva/commons/core/paths/UriWrapper.java
```diff
@@ -146,8 +158,10 @@ public String getFilename() {
 
     public UriWrapper addQueryParameter(String param, String value) {
```
torarnet, line 159: Should we encode in `fromUri `as well or only in `addQueryParameter`?

torarnet: `fromUri(URI/String uri, boolean shouldEncode)?`

torarnet (approved):

truhacevkir (approved):

PR #509, opened 2024-08-26
throw on unhandled token responses

sondrev:
Current issue: AuthorizedBackendClient sends header "Authorization: Bearer null" if it recives a non 200 from cognito /token


auth/src/main/java/no/unit/nva/auth/AuthorizedBackendClient.java
```diff
@@ -123,14 +123,31 @@ private String createBearerToken(String accessToken) {
...
             .map(HttpResponse::body)
             .map(JSON.std::mapFrom)
             .map(json -> json.get(JWT_TOKEN_FIELD))
+            .map(this::assertFieldIsPresent)
             .map(Objects::toString)
             .map(this::createBearerToken)
             .orElseThrow();
     }
 
+    private Object assertFieldIsPresent(Object o) {
```
monaullah, line 136: `Object object`

auth/src/main/java/no/unit/nva/auth/AuthorizedBackendClient.java
```diff
@@ -123,14 +123,31 @@ private String createBearerToken(String accessToken) {
...
+        if (isNull(o)) {
+            throw new IllegalStateException("Received token response without token");
+        }
+        return o;
+    }
+
+    private HttpResponse<String> assertResponseIsOk(HttpResponse<String> response) {
+        if (response.statusCode() < 200 || response.statusCode() >= 300) {
```
monaullah, line 144: How about:
```
private static final Set<Integer> OK_STATUSES = Set.of(HttpURLConnection.HTTP_OK, HttpURLConnection.HTTP_CREATED);


if (!OK_STATUSES.contains(response.statusCode())) {
     throw UnhandledHttpResponseException.fromHttpResponse(response);
}
```

We do not expect any other OK-response codes, do we?

auth/src/main/java/no/unit/nva/auth/UnhandledHttpResponseException.java
```diff
@@ -0,0 +1,16 @@
+package no.unit.nva.auth;
+
+import java.net.http.HttpResponse;
+
+public class UnhandledHttpResponseException extends RuntimeException {
+
+    public static final String UNHANDLED_HTTP_RESPONSE_MESSAGE = "Got unknown http response: %s %s";
```
monaullah, line 7: `unexpected` instead of `unknown`?

truhacevkir (approved):

monaullah (approved):

PR #510, opened 2024-08-26
Np 46577 AuthorizedBackendClient, only refresh token when expired

monaullah:
Experiencing issues during batch scan, Cognito responding with 429 (Too many requests) when requesting tokens. To reduce traffic:
- Use `CachedJwtProvider` in `AuthorizedBackendClient`, only refresh token if expired
- Moved [newly implemented error handling](https://github.com/BIBSYSDEV/nva-commons/pull/509) (by @sondrev) from `AuthorizedBackendClient` to `CognitoAuthenticator`.

### DIFF
--- a/auth/src/main/java/no/unit/nva/auth/AuthorizedBackendClient.java
+++ b/auth/src/main/java/no/unit/nva/auth/AuthorizedBackendClient.java
@@ -1,36 +1,22 @@
 package no.unit.nva.auth;
 
 import static java.util.Objects.isNull;
-import static nva.commons.core.attempt.Try.attempt;
-import static software.amazon.awssdk.http.HttpStatusCode.OK;
-import com.fasterxml.jackson.jr.ob.JSON;
 import java.io.IOException;
-import java.net.URI;
 import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.net.http.HttpRequest.Builder;
 import java.net.http.HttpResponse;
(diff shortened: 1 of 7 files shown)


auth/src/main/java/no/unit/nva/auth/AuthorizedBackendClient.java
```diff
@@ -68,92 +57,32 @@ public static AuthorizedBackendClient prepareWithBearerTokenAndCredentials(HttpC
...
+        refreshTokenIfExpired();
         var authorizedRequest = request.setHeader(AUTHORIZATION_HEADER, bearerToken).build();
         return httpClient.send(authorizedRequest, responseBodyHandler);
     }
 
     public <T> CompletableFuture<HttpResponse<T>> sendAsync(Builder request,
                                                             BodyHandler<T> responseBodyHandler) {
-        refreshToken();
+        refreshTokenIfExpired();
         var authorizedRequest = request.setHeader(AUTHORIZATION_HEADER, bearerToken).build();
```
sondrev, line 70: `var authorizedRequest = request.setHeader(AUTHORIZATION_HEADER, createBearerToken(cachedJwtProvider.getValue().getToken())).build()`
Then we can remove bearerTokenIsNotInjectedDirectly and instead mock needed bearer value by using the cachedJwtProvider

brinxmat (dismissed):

PR #515, opened 2024-09-13
Use wiremock jetty 12

brinxmat:
- DDB local uses jetty v12, which conflicts in deps with Wiremock
- Use wiremock-jetty-12 to avoid this issue


truhacevkir (approved):
Builds on my machine :)

PR #526, opened 2025-01-24
adding addition methods to identity service client

truhacevkir:
Adding methods for fetching customer by id and listing all customers to IdentityServiceClient, so there is noe need to implement "complex configured" http client each time we need to make authenticated request to identity-service.

### DIFF
--- a/clients/src/main/java/no/unit/nva/clients/GetCustomerResponse.java
+++ b/clients/src/main/java/no/unit/nva/clients/GetCustomerResponse.java
@@ -2,11 +2,17 @@ package no.unit.nva.clients;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.net.URI;
+import java.util.List;
 import java.util.UUID;
 import no.unit.nva.commons.json.JsonSerializable;
 
 public record GetCustomerResponse(@JsonProperty("id") URI id, UUID identifier, String name, String displayName,
-                                  String shortName,
-                                  URI cristinId) implements JsonSerializable {
+                                  String shortName, URI cristinId, String publicationWorkflow, boolean nviInstitution,
+                                  boolean rboInstitution, boolean generalSupportEnabled,
+                                  List<String> allowFileUploadForTypes, RightsRetentionStrategy rightsRetentionStrategy)
+    implements JsonSerializable {
 
+    public record RightsRetentionStrategy(String type, URI id) {
+
+    }
 }
--- a/clients/src/main/java/no/unit/nva/clients/IdentityServiceClient.java
+++ b/clients/src/main/java/no/unit/nva/clients/IdentityServiceClient.java
@@ -14,6 +14,7 @@ import java.net.http.HttpResponse;
 import java.util.concurrent.Callable;
 import no.unit.nva.auth.AuthorizedBackendClient;
 import no.unit.nva.auth.CognitoCredentials;
+import nva.commons.apigateway.exceptions.ApiGatewayException;
 import nva.commons.apigateway.exceptions.NotFoundException;
 import nva.commons.core.Environment;
 import nva.commons.core.JacocoGenerated;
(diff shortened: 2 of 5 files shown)


clients/src/main/java/no/unit/nva/clients/GetCustomerResponse.java
```diff
@@ -2,11 +2,17 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.net.URI;
+import java.util.List;
 import java.util.UUID;
 import no.unit.nva.commons.json.JsonSerializable;
 
 public record GetCustomerResponse(@JsonProperty("id") URI id, UUID identifier, String name, String displayName,
```
brinxmat, line 9: Can we rename this class? "GetCustomerResponse" has unfortunate connotations. CustomerDto or Customer?

clients/src/main/java/no/unit/nva/clients/IdentityServiceClient.java
```diff
@@ -95,6 +96,31 @@ public GetCustomerResponse getCustomerByCristinId(URI topLevelOrgCristinId) thro
                    .orElseThrow(this::handleFailure);
     }
 
+    public GetCustomerResponse getCustomerById(URI customerId) throws NotFoundException {
+        var request = HttpRequest.newBuilder()
+                          .GET()
+                          .uri(customerId);
+        return attempt(getHttpResponseCallable(request))
+                   .map(this::validateResponse)
+                   .map(r -> mapResponse(GetCustomerResponse.class, r))
```
brinxmat, line 105: r?

clients/src/main/java/no/unit/nva/clients/IdentityServiceClient.java
```diff
@@ -95,6 +96,31 @@ public GetCustomerResponse getCustomerByCristinId(URI topLevelOrgCristinId) thro
                    .orElseThrow(this::handleFailure);
     }
 
+    public GetCustomerResponse getCustomerById(URI customerId) throws NotFoundException {
+        var request = HttpRequest.newBuilder()
+                          .GET()
+                          .uri(customerId);
```
brinxmat, line 102: This can be extracted, and add relevant headers (e.g. Accept, possibly Accept-Encoding) since there are some defaults that Java inserts (actually, check these by adding a breakpoint and get back to me)

clients/src/main/java/no/unit/nva/clients/IdentityServiceClient.java
```diff
@@ -95,6 +96,31 @@ public GetCustomerResponse getCustomerByCristinId(URI topLevelOrgCristinId) thro
...
+        return attempt(getHttpResponseCallable(request))
+                   .map(this::validateResponse)
+                   .map(r -> mapResponse(GetCustomerResponse.class, r))
+                   .orElseThrow(this::handleFailure);
+    }
+
+    public CustomerList getAllCustomers() throws ApiGatewayException {
+        var request = HttpRequest.newBuilder()
+                          .GET()
+                          .uri(constructListCustomerUri());
```
brinxmat, line 112: Use the extracted method

clients/src/test/java/no/unit/nva/clients/IdentityServiceClientTest.java
```diff
@@ -215,6 +217,73 @@ void shouldThrowNotFoundWhenCustomerNotFound() throws IOException, InterruptedEx
...
     }
 
+    @Test
+    void shouldReturnCustomerByIdWhenRequested() throws NotFoundException, IOException, InterruptedException {
+        var customerId = randomUri();
+        var expectedCustomer = createCustomerWithCristinId(customerId);
+        var request = HttpRequest.newBuilder()
+                          .GET()
+                          .uri(customerId)
+                          .build();
```
brinxmat, line 227: Extract

brinxmat (changes requested):

clients/src/main/java/no/unit/nva/clients/IdentityServiceClient.java
truhacevkir: I think all the headers already are getting set at getHttpResponseCallable(request) method, like accept header.

torbjokv (approved):
One "critical" question we should think about, otherwise good :-)

PR #527, opened 2025-02-21
creating cristin client for fetching person and organization

truhacevkir:
Creating common client for fetching Cristin Person and Cristin organization by identifier and id.

Why?

Today we are fetching orgs and person at least in following processes and each service has its own implementation of doing it:

- ScopusImport
- Expansion
- Nvi
- IdentityService

Now we will fetch person and organization one more place - LogService. 
I think this is a suited time to create common service and in future remove all "duplicated" implementations all over the projects.


clients/src/main/java/no/unit/nva/clients/cristin/CristinClient.java
```diff
@@ -0,0 +1,86 @@
...
+    private final HttpClient httpClient;
+
+    public CristinClient(HttpClient httpClient) {
+        this.httpClient = httpClient;
+    }
+
+    @JacocoGenerated
+    public static CristinClient prepareNotAuthorizedClient() {
+        return new CristinClient(HttpClient.newBuilder().build());
+    }
```
truhacevkir, line 37: Once it will be needed, we can add authenticated client, which serves more data about person.

clients/src/main/java/no/unit/nva/clients/cristin/CristinClient.java
```diff
@@ -0,0 +1,86 @@
...
+    public static final String ORGANIZATION = "organization";
+    private static final Logger LOGGER = LoggerFactory.getLogger(CristinClient.class);
+    private static final String API_HOST = new Environment().readEnv("API_HOST");
+    private final HttpClient httpClient;
+
+    public CristinClient(HttpClient httpClient) {
+        this.httpClient = httpClient;
+    }
+
+    @JacocoGenerated
```
LarsV123, line 34: Does this need to be excluded from tests? Can we test it here?

clients/src/main/java/no/unit/nva/clients/cristin/Affiliation.java
LarsV123, line 1: Should this be named `AffiliationDto` to match the other files here?

clients/src/main/java/no/unit/nva/clients/cristin/Role.java
LarsV123, line 1: Same as for `Affiliation`, is this not a "DTO" also?

clients/src/main/java/no/unit/nva/clients/cristin/Affiliation.java
truhacevkir: I think it is fine to name it Affiliation and Role, cause we do not have other affiliations and roles and mappings between business/dao/dto objects.

I think it is not necessary, in addition I think it is not so clean when we end up with methods like:
CristinOrganizationDto.affiliationDto().roleDto(); Enough with dto in the first method in chain I think.

PR #530, opened 2025-02-28
support for viewing scope

ketilaa:


apigateway/src/main/java/nva/commons/apigateway/RequestInfo.java
```diff
@@ -414,6 +431,10 @@ private Optional<String> fetchUserName() {
         return fetchUserInfo().map(CognitoUserInfo::getUserName);
     }
 
+    private Optional<ViewingScope> fetchViewingScope() {
+        return fetchUserInfo().map(u -> ViewingScope.from(u.getViewingScopeIncluded(), u.getViewingScopeExcluded()));
```
truhacevkir, line 435: maybe userInfo instead of u

truhacevkir (dismissed):

apigateway/src/main/java/nva/commons/apigateway/RequestInfo.java
ketilaa: fixed

truhacevkir (approved):

PR #534, opened 2025-03-24
NP-48779 Add access right for channel claims

joachimjorgensen:


brinxmat (approved):

apigateway/src/main/java/nva/commons/apigateway/AccessRight.java
```diff
@@ -67,6 +67,7 @@ public enum AccessRight {
     SUPPORT("support"),
     MANAGE_KBS("kbs"),
     MANAGE_RESOURCE_FILES("resource-files"),
+    MANAGE_CHANNEL_CLAIMS("channel-claims"),
```
torbjokv, line 70: maybe we should start documenting what these mean in addition to the previous names stuff? MANAGE_CHANNEL_CLAIMS is that for sikt admins or people at the institution

torbjokv (approved):

PR #538, opened 2025-04-23
fix(auth): AuthorizedBackendClient should validate hostname #NP-48701

torbjokv:


auth/src/main/java/no/unit/nva/auth/AuthorizedBackendClient.java
```diff
@@ -61,9 +63,21 @@ public <T> HttpResponse<T> send(HttpRequest.Builder request, BodyHandler<T> resp
         throws IOException, InterruptedException {
         refreshTokenIfExpired();
         var authorizedRequest = request.setHeader(AUTHORIZATION_HEADER, bearerToken).build();
+
+        if (!hasValidBackendHost(authorizedRequest)) {
```
brinxmat, line 67: is this always negated? `hasInvalid…`

brinxmat (changes requested):

auth/src/main/java/no/unit/nva/auth/AuthorizedBackendClient.java
torbjokv: I just dont like negated methods in general, but maybe its just my head that's like that. I can turn it around :-)

joachimjorgensen (dismissed):

brinxmat (dismissed):

PR #541, opened 2025-05-07
adding id field to channel claim dto

truhacevkir:
ChannelClaim from identity-service contains id. We will reuse it in publication-api, instead of constructing it based on publication-channel identifier.


joachimjorgensen (approved):

brinxmat (changes requested):
Can you please ensure that you have bumped jackson and other vulns.

brinxmat (approved):

PR #599, opened 2026-04-21
build: Use shared Gradle plugins

LarsV123:
Changes:
- Replaces Gradle plugin configuration with a reference to shared plugin (´nva-gradle-plugins`)
- Removes dead code related to previous attempts at publishing a shared Gradle plugin from this repo
- Refactors the project to only use `build-logic` for Gradle plugins and removes the `buildSrc` directory


build-logic/src/main/groovy/nvacommons.root-module-conventions.gradle
```diff
@@ -0,0 +1,12 @@
...
+}
+
+nva {
+    dependencyAnalysis {
+        enforced = true
+    }
+    spotless {
+        enabled = false
+    }
+}
```
LarsV123, line 12: The repo did not have `spotless` properly configured and running it now reformats hundreds of files, so I'm leaving it disabled here. We can enable it and reformat in a separate PR.

torbjokv (approved):

